PR #25, opened 2026-04-01
fix: add --dry-run flag to install command

### LINKED ISSUE #23: Add --dry-run flag to install command
## Description

Add a `--dry-run` flag to `homebutler install <app>` that shows the generated `docker-compose.yml` without actually installing.

## Expected behavior

```bash
homebutler install pi-hole --dry-run
```

Should output the rendered docker-compose.yml to stdout and exit — no Docker pull, no container creation.

## Implementation hints

- Add `--dry-run` flag in `cmd/install.go` (`newInstallCmd`)
- In `runInstallApp()`, after compose rendering but before `docker compose up`, check the flag
- If dry-run: print compose content to stdout and return
- Tests: verify dry-run doesn't create files or containers

## Difficulty
Beginner — one flag + one conditional branch.

Deepak8858:
This PR adds a `--dry-run` flag to the `install` command, allowing users to see what would happen before actually performing the installation. Fixes #23.

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -37,6 +37,7 @@ Usage:
 
 	installCmd.Flags().String("port", "", "Custom host port")
 	installCmd.Flags().String("media", "", "Media directory to mount (jellyfin)")
+	installCmd.Flags().Bool("dry-run", false, "Do not actually install, just show what would happen")
 
 	installCmd.AddCommand(
 		newInstallListCmd(),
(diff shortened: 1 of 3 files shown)


Higangssh (changes requested):
@Deepak8858 Thanks for picking this up! Good start. A few things to address:

**1) Dry-run should not create files on disk**
Currently `Install()` still creates the app directory and writes `docker-compose.yml` before the dry-run check. The compose file should be rendered to stdout instead of written to disk.

Suggested approach:
```go
if opts.DryRun {
    // Render to stdout instead of writing to file
    tmpl.Execute(os.Stdout, ctx)
    return nil
}
```

**2) Show the compose content**
Users expect `--dry-run` to show *what* would be installed. Currently it only prints "Would run: docker compose up". Please print the rendered compose YAML so users can inspect it.

**3) Missing tests**
Please add at least:
- Test that dry-run returns nil error
- Test that dry-run does not create files in the app directory
- Test that compose content is rendered correctly

Take a look at existing tests in `internal/install/install_test.go` for the pattern.

Overall direction is right — just needs these fixes before merge. Let me know if you have questions!

Higangssh (approved):

PR #35, opened 2026-08-15
test: add fixture tests for inventory port parsing

### LINKED ISSUE #29: Add fixture tests for inventory port parsing
## Description
Add fixture-based tests for inventory port parsing.

Inventory output depends on parsing platform-specific command output such as Linux `ss` and macOS `lsof`. Fixture tests will make this safer to change and easier for new contributors to understand.

## Scope
- Add test fixtures under `internal/inventory/testdata/`
- Cover at least one Linux `ss` output example
- Cover at least one macOS `lsof` output example
- Test that parsed ports include protocol, port number, process name, and exposure/public-local classification where applicable
- Do not change runtime command execution in this issue unless required for testability

## Suggested fixtures
```text
internal/inventory/testdata/ss-linux.txt
internal/inventory/testdata/lsof-macos.txt
```

## Implementation hints
- Look for existing port parsing code in `internal/inventory`
- If parsing is currently embedded too deeply, extract a small pure function that accepts command output as a string
- Keep fixtures generic: use names like `app-api`, `metrics-agent`, or `demo-service`
- Do not include real hostnames, private user paths, or personal IP addresses

## Done when
- Fixture tests cover Linux `ss` parsing
- Fixture tests cover macOS `lsof` parsing
- Tests use only dummy data
- `go test ./internal/inventory` passes
- `go test ./...` passes

lenny-ts:
Export ParseLinuxOutput and ParseDarwinOutput from internal/ports so the inventory package can drive fixture-based tests without spawning ss/lsof. Runtime command execution is unchanged.

Add testdata fixtures for Linux ss and macOS lsof output using generic dummy processes (app-api, demo-service, metrics-agent, edge-router, ssh-relay). Tests assert protocol, port number, process name, address, and public/local exposure classification.

Closes #29

### DIFF
--- a/internal/inventory/portparse_fixture_test.go
+++ b/internal/inventory/portparse_fixture_test.go
@@ -1,0 +1,106 @@
+package inventory
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/Higangssh/homebutler/internal/ports"
+)
+
+// portExpect captures the fields the fixture tests assert on for a single port.
+type portExpect struct {
+	port    string
+	process string
+	address string
+	public  bool
+}
+
+func TestParseLinuxSSFixture(t *testing.T) {
+	out := mustReadFixture(t, "ss-linux.txt")
+	parsed := ports.ParseLinuxOutput(string(out))
+
+	if len(parsed) != 5 {
+		t.Fatalf("expected 5 parsed ports from ss fixture, got %d", len(parsed))
+	}
+
+	expects := []portExpect{
+		{port: "8080", process: "app-api", address: "0.0.0.0", public: true},
+		{port: "443", process: "demo-service", address: "[::]", public: true},
+		{port: "9090", process: "metrics-agent", address: "127.0.0.1", public: false},
+		{port: "8443", process: "edge-router", address: "0.0.0.0", public: true},
+		{port: "22", process: "ssh-relay", address: "0.0.0.0", public: true},
+	}
+
+	for _, e := range expects {
+		p, ok := findPortByNumber(parsed, e.port)
+		if !ok {
+			t.Errorf("port %s: not found in parsed output", e.port)
+			continue
+		}
+		assertPort(t, e, p)
+	}
+}
+
+func TestParseDarwinLsofFixture(t *testing.T) {
+	out := mustReadFixture(t, "lsof-macos.txt")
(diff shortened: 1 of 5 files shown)


Higangssh:
Merged — thanks for this.

Before anything else, a correction I owe you on the issue. #29 told you to look for the parsing code in `internal/inventory`, and to extract a pure function if it was buried too deep. Neither was actually true: parsing lives in `internal/ports`, and `parseLinuxOutput` / `parseDarwinOutput` were already pure with tests in `ports_test.go`. I wrote that issue from memory without opening the package first. So if the layout felt like it didn't match the description while you were working, that's why — and you navigated around a wrong premise cleanly.

Things I liked:

- The dedup comment (`5 input lines, but the IPv4/IPv6 *:8080 pair for app-api is deduped to 1`). That tells me you traced `parseDarwinOutput` rather than adjusting a number until the test went green. It's the exact spot I'd expect someone to get wrong.
- Looking ports up by number instead of by index. The older tests in `ports_test.go` use `ports[0]` / `ports[2]`, which quietly couples them to parser ordering; yours don't. I'd like to move the existing ones in that direction.
- Dummy names from the issue, no real data. Appreciated.

**One thing your fixtures surfaced that neither of us caught.** The `ss` fixture carries `pid=1001` and friends, and `ParseLinuxOutput` throws every one of them away — it never sets `PortInfo.PID`, while `ParseDarwinOutput` does (`ports.go:86` vs `ports.go:124`). Since `PID` is a serialized field, `homebutler ports --json` reports a pid on macOS and an empty one on Linux, which is backwards from where most of our users actually are.

Your `portExpect` has no `pid` field, so the test passes right over it. That's not a knock on the PR — the bug predates it, which is why I didn't hold the merge for it. But a fixture whose job is pinning parser output is exactly where it should get caught.

I'll open a separate issue for it. **If you'd like that one too, say the word and it's yours** — the fix is small, and adding `pid` to `portExpect` would make these fixtures assert it properly.

Two other things I'm tracking, neither of them yours to fix: `isPublicBind` exists twice with different implementations (`internal/inventory/render.go:249` and `internal/doctor/doctor.go:278`), and parser tests now live in two packages, which is what the exports bought us. I'd rather consolidate both when I do the JSON output schema work than churn a PR that's already good.

Thanks again — solid first contribution.

Issue #28, opened 2026-04-26
Add --filter exposed to inventory scan


Higangssh:
## Description
Add a filter option to `homebutler inventory scan` so users can quickly see only exposed/public services.

This is useful when someone wants to answer: "What is reachable from outside my machine/network?" without reading the full inventory tree.

## Expected behavior
```bash
$ homebutler inventory scan --filter exposed
```

Example output:
```text
🏠 Home Network
   Server  demo-lab

🌐 Exposed Ports
   ├─ :8080/tcp · app-api
   └─ :8443/tcp · dashboard
```

## Scope
- Add a `--filter` flag to `inventory scan`
- Support at least `--filter exposed`
- Only show public/exposed app ports and their related containers/services
- Keep default `inventory scan` behavior unchanged
- Return a clear error for unsupported filter values

## Implementation hints
- Start in `cmd/inventory.go`
- Filtering can happen after the inventory snapshot is collected
- Reuse the existing inventory renderers where possible
- Avoid changing Docker or port scanning behavior in this issue

## Done when
- `homebutler inventory scan --filter exposed` works
- Unsupported filters return a helpful error
- Existing inventory output remains unchanged without the flag
- Tests cover the exposed filter behavior
- `go test ./...` passes

Higangssh:
@lenny-ts — this one's yours, following on from #40.

Small mechanical thing: GitHub only lets me set an assignee if the person has push access or has commented on the issue itself. Your ask was on the PR, so the assignment silently no-ops here. Drop any reply below and I'll set it properly — it's a GitHub quirk, not a hoop I'm making you jump through.

Consider it claimed either way, and no rush, same as last time.

Two pointers so you don't have to go looking:

- The exposure predicate you want is `isPublicBind` in `internal/inventory/render.go`. There's a second copy in `internal/doctor/doctor.go` that #37 will merge into one — ignore that, use the inventory one and don't wait on #37.
- `splitPorts` in the same file already separates app ports from system ports using the Docker port links, which is most of the grouping the filter output needs.

One thing worth knowing before you start: `inventory scan` is local-only. It doesn't call `maybeRouteRemote()`, and `runLocalCommand` in `cmd/helpers.go` has no `inventory` case, so `--server` and `--all` don't apply. The filter is purely a local render-side concern, which keeps this smaller than it might look.

lenny-ts:
reply :)

PR #40, opened 2026-08-17
fix: Populate PID in ParseLinuxOutput

### LINKED ISSUE #36: ports: ParseLinuxOutput drops the PID that ss already reports
## Description

`PortInfo` has a `PID` field that is serialized to JSON:

```go
type PortInfo struct {
	Protocol string `json:"protocol"`
	Address  string `json:"address"`
	Port     string `json:"port"`
	PID      string `json:"pid,omitempty"`
	Process  string `json:"process,omitempty"`
}
```

`ParseDarwinOutput` populates it (`internal/ports/ports.go:86`). `ParseLinuxOutput` never does (`internal/ports/ports.go:124`), even though the PID is sitting right there in the `ss` output it already parses:

```
LISTEN  0  128  0.0.0.0:8080  0.0.0.0:*  users:(("app-api",pid=1001,fd=3))
                                                            ^^^^^^^^^
```

The parser reaches into that same field to pull out the process name and stops there.

**Effect:** `homebutler ports --json`, `inventory scan --json`, and the MCP tools return a `pid` on macOS and omit it on Linux — the platform most homelab users are actually on. Anything downstream that wants to correlate a listener with a process has to shell out again on Linux.

Surfaced while reviewing #35, whose `ss` fixtures carry `pid=1001` and friends. The bug predates that PR.

## Scope

- Extract the PID in `ParseLinuxOutput` alongside the process name
- Keep it a `string`, matching the Darwin path
- Leave it empty when `ss` reports no process info (unprivileged runs) — `TestParseLinuxOutput_NoProcessInfo` covers that shape
- Add a `pid` field to `portExpect` in `internal/inventory/portparse_fixture_test.go` and assert it against the fixtures

## Notes

`ss` can list more than one process per socket:

```
users:(("sshd",pid=1234,fd=3),("sshd",pid=5678,fd=4))
```

`PortInfo` holds a single PID, so take the first entry — consistent with how the process name is already handled.

## Done when

- `ParseLinuxOutput` sets `PID` when `ss` reports one
- The fixture tests in `internal/inventory` assert `pid` for every fixture row
- Unprivileged output with no process column still parses with empty `PID` and `Process`
- `go test ./...` passes

---

@lenny-ts — this is the follow-up I mentioned on #35. Yours if you want it, no obligation.

lenny-ts:
resolve #36

### DIFF
--- a/internal/inventory/portparse_fixture_test.go
+++ b/internal/inventory/portparse_fixture_test.go
@@ -12,6 +12,7 @@ import (
 type portExpect struct {
 	port    string
 	process string
+	pid     string
 	address string
 	public  bool
 }
@@ -25,11 +26,11 @@ func TestParseLinuxSSFixture(t *testing.T) {
 	}
 
 	expects := []portExpect{
-		{port: "8080", process: "app-api", address: "0.0.0.0", public: true},
-		{port: "443", process: "demo-service", address: "[::]", public: true},
-		{port: "9090", process: "metrics-agent", address: "127.0.0.1", public: false},
-		{port: "8443", process: "edge-router", address: "0.0.0.0", public: true},
-		{port: "22", process: "ssh-relay", address: "0.0.0.0", public: true},
+		{port: "8080", process: "app-api", pid: "1001", address: "0.0.0.0", public: true},
+		{port: "443", process: "demo-service", pid: "1002", address: "[::]", public: true},
+		{port: "9090", process: "metrics-agent", pid: "1003", address: "127.0.0.1", public: false},
+		{port: "8443", process: "edge-router", pid: "1004", address: "0.0.0.0", public: true},
+		{port: "22", process: "ssh-relay", pid: "1005", address: "0.0.0.0", public: true},
 	}
 
 	for _, e := range expects {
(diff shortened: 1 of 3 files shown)


Higangssh:
Merged — nice work, and fast.

`strings.Index` finding the first `pid=` is the right call, and I'm glad you pinned it with `TestParseLinuxOutput_MultipleProcesses` rather than leaving it implicit. The comment on that test explains *why* a single PID is correct rather than just what the test does, which is the part people usually skip.

I poked at a few shapes beyond what the tests cover, and the digit scan holds up:

| input | process | pid |
|---|---|---|
| `users:(("nginx",fd=3))` — no pid at all | `nginx` | `""` |
| `users:(("my,proc",pid=7,fd=3))` — comma in name | `my,proc` | `7` |
| `*:22` instead of `0.0.0.0:22` | `sshd` | `42` |
| `pid=4194304` | `x` | `4194304` |

The one that could have gone wrong is a process name containing a space, since `strings.Fields` splits the users:(...) blob across two entries. Both the name and the PID read from `fields[5]`, so they come back empty together instead of one silently disagreeing with the other. That limitation predates this PR, and keeping the two in lockstep is the behaviour I'd want.

One thing worth naming that isn't in the diff: `format.Ports` appends `/PID` whenever a PID is present, so the human-readable `homebutler ports` output on Linux now reads `nginx/5678` where it used to read `nginx`. That's the right outcome — macOS already printed it that way and the two platforms should agree — but it is a visible change, so I wrote it into the squash commit message and it'll go in the changelog. I checked that inventory's port deduplication doesn't key on PID, so grouping there is unaffected.

Two small things I fixed on merge rather than sending back, just so you know for next time: the repo keeps the conventional-commit subject lowercase after the colon (`fix: populate ...`), and commit bodies here tend to carry the why. I filled that in from the issue.

Thanks for turning this around so quickly. If you want another, #28 (`inventory scan --filter exposed`) is small, well-specified, and unclaimed — no pressure either way.

lenny-ts:
Hi @Higangssh  Thanks for the careful review — and for the merge. 
On #28 — I'd be happy to take it. Assign it to me and I'll pick it up when I have a focused window to give it proper attention. I genuinely enjoy contributing to other people's projects, and the way you review (poking at shapes beyond the tests, naming the things that aren't in the diff) makes the work better. No rush on my end either.

PR #48, opened 2026-08-21
feat: cap incident history so the watch directory stops growing forever

Higangssh:
> **Stacked on #46.** Based on `fix/watch-check-skipped-targets` because both touch
> `CheckTargets`. Review that one first; GitHub retargets this to `main` when it merges.
> The diff shown here is this change only.

## The problem

Nothing ever deletes an incident. `SaveIncident` writes one JSON file per incident and
there is no prune, rotate, or retention anywhere in `internal/watch`.

Each file embeds the captured logs — `journalctl -u <unit> -n 100` for systemd,
`docker logs --tail 100` for containers — so the directory grows fastest exactly when
something is wrong. A service in a restart loop at 30s intervals writes roughly 2,880
incidents a day, each tens of kilobytes, none of them ever collected.

Flapping detection does not bound this. `FlappingConfig.Check` classifies an incident once
it already exists; it never suppresses the write. So the case that produces the most files
is the one flapping was supposed to be about.

Noticed while looking at what `watch` would do on a small host — 16GB disk, 0.9GB RAM —
where a runaway incident directory is not a theoretical concern.

## The fix

`SaveIncident` enforces a cap after each write: keep the newest N, delete the rest.

Pruning lives on the write path rather than at the call sites. There are five writers —
the three monitors, `CheckTargets`, and the enrichment re-save in `cmd/watch.go` — and a
cap each caller had to remember to apply is a cap one of them eventually forgets. The
monitors carry a `Keep` field; `CheckTargets` and `SaveIncident` take it as a parameter.

## Configuration

```yaml
watch:
  retention:
    max_incidents: 200   # default; -1 keeps everything
```

Resolved like the neighbouring watch settings: `config.yaml` wins over
`watch/config.json`.

The sentinel handling is the part worth reviewing. `0` means **not configured** and takes
the default; `-1` is the explicit opt-out. If `0` meant unlimited, every `config.json`
written before this setting existed — they have no `retention` key, so it decodes to the
zero value — would silently read as "keep everything", which is the exact behaviour this
PR exists to end. `Normalize()` folds `-1` into the `0` that `PruneIncidents` treats as
unlimited, so the internal contract stays simple.

`config validate` needed `retention` and `max_incidents` added to its key allowlist —
`WatchRuntimeConfig` has a custom unmarshaler, so `KnownFields` does not reach into the
`watch:` subtree and a new block would otherwise be reported as an unknown key.

## Two deliberate non-behaviours

**Unparseable incident files are never deleted.** `ListIncidents` skips files it cannot
read or unmarshal, so pruning never selects them. They can accumulate. Refusing to delete
a file we do not understand seemed the safer half of that trade; the alternative is
deleting on a parse error, which is a worse failure mode than leaking a few bytes.

**A failed prune does not fail the save.** The incident is already on disk by then.
Returning an error would tell the caller the incident was lost when it was not, so the
prune error goes to stderr as a warning.

## Also here

`watch check` now calls `loadConfig()`, as `watch start` has since #31. Without it, `check`
would read retention only from `watch/config.json` while `start` read it from
`config.yaml`, and the two commands would enforce different caps on the same directory.

## Tests

`internal/watch`:
- `TestPruneIncidents_KeepsNewestAndDeletesOldest` — saved oldest-first so file order
  cannot be what makes it pass.
- `TestPruneIncidents_UnderCapIsNoop`, `TestPruneIncidents_MissingDirIsNoop`
- `TestPruneIncidents_NonPositiveKeepsEverything` — both `0` and `-1`.
- `TestSaveIncident_EnforcesCap` — the write path caps on its own, without an explicit
  prune call. This is the one that would catch a future monitor bypassing the cap.
- `TestRetentionConfig_Normalize`, `TestLoadWatchConfig_AppliesRetentionDefault` — the
  legacy-config case, written against a `config.json` with no `retention` key.

`internal/config`:
- `TestLoadWatchRetention`, `TestLoadWatchRetentionDefaultsWhenUnset`,
  `TestLoadWatchRetentionNegativeMeansUnlimited`
- `TestValidateAcceptsWatchRetention`, `TestValidateUnknownWatchRetentionKey`

`go build ./...` and `go vet ./...` clean. `internal/watch`, `internal/config`, and `cmd`
pass.

README documents the setting under the existing watch config block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,8 @@ watch:
     short_threshold: 3
     long_window: 24h
     long_threshold: 5
+  retention:
+    max_incidents: 200
 
 alerts:
   cpu: 90
@@ -346,6 +348,7 @@ Legacy `~/.homebutler/watch/config.json` is still read as a fallback for watch-s
 - `watch.notify_on: off` — disable watch notifications without removing provider config
 - `watch.cooldown: 5m` — suppress duplicate notifications for the same event fingerprint during the cooldown window
 - `watch.flapping` — optional advanced tuning for restart-loop detection
+- `watch.retention.max_incidents: 200` — how many incidents to keep on disk, newest first. Each incident stores up to 100 captured log lines, so the directory grows fastest exactly when a service is restarting in a loop. Set `-1` to keep everything.
 
 These settings can also be written under a `watch.notify:` block, which is the
 canonical form:
--- a/cmd/watch.go
+++ b/cmd/watch.go
@@ -275,6 +275,11 @@ func newWatchCheckCmd() *cobra.Command {
 		Use:   "check",
 		Short: "Run a one-shot restart check on all watched containers",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Loaded so config.yaml's watch settings reach this command too;
+			// without it check and start would disagree about retention.
+			if err := loadConfig(); err != nil {
+				return err
+			}
 			dir, err := watch.WatchDir()
 			if err != nil {
 				return err
@@ -288,7 +293,7 @@ func newWatchCheckCmd() *cobra.Command {
 				return nil
 			}
 
-			result, err := watch.CheckTargets(dir)
+			result, err := watch.CheckTargets(dir, resolveIncidentCap(dir))
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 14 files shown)


Higangssh:
Superseded by #50. This PR was based on `fix/watch-check-skipped-targets` (#46) and GitHub closed it when that branch was deleted on merge — a closed PR cannot be reopened or retargeted once its base is gone. The branch was rebased onto `main` and reopened as #50 with the same content.

PR #51, opened 2026-08-21
feat: add --filter exposed to inventory scan

### LINKED ISSUE #28: Add --filter exposed to inventory scan

lenny-ts:
## What does this PR do?

Adds a `--filter` flag to `homebutler inventory scan`, supporting `--filter exposed`,
so users can quickly answer *"what is reachable from outside my machine/network?"*
without reading the full inventory tree.

```bash
homebutler inventory scan --filter exposed
```

```text
🏠 Home Network
   Server  demo-lab

🌐 Exposed Ports
   ├─ :8080/tcp · app-api
   └─ :8443/tcp · dashboard
```

### Implementation notes

- Filtering is purely render-side (`internal/inventory/render.go`): the snapshot is
  collected as usual, then rendered through the new `RenderTreeFiltered`. No changes
  to Docker or port scanning behavior.
- Reuses the existing exposure predicate (`ports.IsPublicBind`) and dedupe logic;
  ports show their Docker container as owner when linked, otherwise the process name.
- Collection warnings are surfaced in the filtered view too, so a failed port scan
  never reads as "(none)".
- Unsupported filter values fail fast with a clear error before any collection:
  `unsupported filter "bogus" (supported: exposed)`
- Default `inventory scan` output is unchanged; JSON output is not filtered
  (documented in README). `inventory show` keeps its current behavior.
- Local-only by design: no `maybeRouteRemote` involvement, matching the issue scope.

### Testing

- 7 new tests in `internal/inventory`: mixed public/local ports with container-linked
  owner, public IPv6 bind, empty result `(none)`, collection warnings rendering,
  unsupported filter error, `SupportedFilters`/`IsSupportedFilter`.
- Verified locally: `gofmt`, `go vet`, `golangci-lint` (0 issues), `go test ./... -race -count=1`,
  `go mod tidy -diff`, plus manual smoke tests of all cases above.
- gosec on the touched packages reports no findings.

Closes #28 

## Checklist

- [x] README updated
```

### CODE REVIEW

**Comment by Higangssh** [comment]
Thanks — this is a clean one.

Reusing `dockerPortLinks` / `dedupeDisplayPorts` / `friendlyPortOwner` rather
than adding a parallel rendering path is exactly right, and `ports.IsPublicBind`
landing in #47 last week made this cheaper than it would otherwise have been.

The part worth calling out is surfacing collection warnings in the filtered
view. A failed port scan rendering as a bare `(none)` would read as "nothing is
exposed" — the one wrong answer this command must never give. That's normally
something I have to ask for in review, so thanks for getting there first, and
for the test that pins it.

Two things:

1. The description still has the `Closes #<issue-number>` placeholder. Making it
   `Closes #28` will close the issue on merge.
2. Non-blocking: `unsupported filter %q (supported: %s)` now lives in both
   `cmd/inventory.go` and `render.go`. The double validation is right — failing
   before collection is the better behaviour — but the two strings will drift
   apart eventually. Happy to take that as a follow-up.

Fix the first and I'll merge. Third one from you now — much appreciated.

**Comment by lenny-ts** [comment]
Updated the description

**Comment by lenny-ts** [comment]
Whenever this merges, I'd be glad to take on another issue. Feel free to assign me anything that fits — I'm enjoying contributing to homebutler.

**Comment by Higangssh** [comment]
Description looks right now, thanks.

Before merging I went back through this more carefully than I did the first time, and three things need to change. All small, all in scope.

**1. `--json` accepts `--filter` and ignores it.** (blocking)

`runInventoryScan` returns at the `jsonOutput` branch before the filter branch is reached, so `homebutler --json inventory scan --filter exposed` prints the full inventory. I know this is deliberate and documented — the README says JSON output is not filtered — but a flag that is accepted and then does nothing is a footgun regardless of where that's written down, and scripts don't read the README. Someone doing

```bash
homebutler --json inventory scan --filter exposed | jq '.ports[].port'
```

gets every loopback listener back and counts it as exposed. That is the one wrong answer this command must never give.

I'm not asking you to filter the JSON — that opens schema questions I don't want in this PR. Reject the combination before collection, next to the filter validation that's already there.

**2. The README qualifier is narrower than it reads.** (blocking)

> Only ports bound to a public interface (`0.0.0.0`, `::`, `*`) are shown; loopback listeners are hidden.

The first half is accurate. The second half implies loopback is the only thing hidden. `nginx` bound to `192.168.1.10:443` is neither a wildcard bind nor a loopback listener, so it's hidden too — while being reachable from every other machine on the network. Someone running a reverse proxy reads `(none)` and stops looking.

That part isn't yours: `IsPublicBind` has been wildcard-only since I consolidated it in #47, and reusing it here was the right call. But the sentence should describe what the filter actually does. Concretely, replace it with:

> Only ports listening on all interfaces (`0.0.0.0`, `::`, `*`) are shown. Anything bound to a specific address is hidden, including loopback and LAN addresses.

That matches the phrasing `doctor` already uses ("listening on all interfaces"), so the two commands describe the same predicate the same way. Whether the predicate itself should widen is a separate question that affects `doctor` too, so I'll open an issue for it rather than push it into this PR.

**3. `(none)` still prints when the scan failed.** (blocking, and my miss)

I praised this last time without checking it. The warning does render, but `len(exposed) == 0` is unconditional, so a failed `ports.List()` gives you:

```text
🌐 Exposed Ports
   (none)

⚠️  Warnings (1)
   └─ ...
```

`(none)` is still the first thing the eye lands on, which is the failure mode we both wanted to avoid. Skip the `(none)` line entirely when `inv.Warnings` is non-empty — I'd rather print nothing under the heading than assert a count we don't have. And extend `TestRenderTreeFiltered_ExposedShowsWarnings` to assert `(none)` is *absent* in that case; right now it only checks that the warning is present, which is why this got through.

So, concretely:

- [ ] `cmd/inventory.go` — error out when `--json` and `--filter` are combined, alongside the existing filter validation
- [ ] `README.md` — swap the qualifier for the sentence above
- [ ] `render.go` — skip `(none)` when `inv.Warnings` is non-empty
- [ ] `inventory_test.go` — assert `(none)` is absent in the warnings case

Fix those four and I'll merge.

**Follow-ups, not this PR — these don't block the merge:**

- `inventory show` advertises "same as scan" in its `Short` but rejects `--filter`, since the flag is only registered on `scan`.
- The sharper version of the drift I raised last time: `RenderTreeFiltered`'s `switch` and `SupportedFilters()` are separate sources of truth. Add a filter to the slice, forget the `case`, and you get `unsupported filter "listening" (supported: exposed, listening)` — an error that contradicts itself. Worth doing as one change together with the `show` flag.

Minor: there's a stray code fence at the end of the description, left over from the edit. Harmless — it just renders an empty block.

On the next issue: yes, and I'll tag you on one this week. If you'd rather stay in this area, the two follow-ups above are yours for the asking.

lenny-ts:
All four fixed and pushed in one commit (`677909b`):

1. `--json` + `--filter` now rejected before collection, with a cmd-level test pinning it
2. README swapped for your sentence verbatim (JSON clause updated to describe the rejection)
3. `(none)` skipped when warnings are present — a failed scan prints no count
4. Warning test extended to assert `(none)` is absent


Stray fence in the description removed. Happy to take both follow-ups — the `show` flag plus the switch/`SupportedFilters()` consolidation as one change.

Issue #61, opened 2026-08-22
hasSecrets misses notify credentials, so configs holding only a bot token are never permission-checked


Higangssh:
## Description

`config.Load` enforces `0600` on the config file when it holds secrets. It decides that with `hasSecrets` (`internal/config/config.go:261`):

```go
func hasSecrets(cfg *Config) bool {
	for _, s := range cfg.Servers {
		if s.Password != "" {
			return true
		}
	}
	return false
}
```

Only `servers[].password` counts. Notify credentials do not.

So a config containing a Telegram bot token, a Slack webhook, or a Discord webhook and no SSH password is never permission-checked. The file can be world-readable and homebutler says nothing. The check exists, it just never runs for the credentials most users actually have — `watch` notifications are set up far more often than SSH passwords.

## The second half

`internal/notify/config.go` gives the secret fields JSON tags:

```go
BotToken   string `yaml:"bot_token" json:"bot_token"`
WebhookURL string `yaml:"webhook_url" json:"webhook_url"`
```

Nothing serializes `*config.Config` today, so nothing leaks right now. But that is a property of the current call sites, not of the types. `report`, `doctor`, and the MCP tools all serialize a fair amount of state, and the surface they serialize keeps growing. A secret that is structurally serializable will eventually be serialized by someone who did not know it was there.

## Scope

- Teach `hasSecrets` about notify credentials — bot token, and each webhook URL
- Tag secret fields `json:"-"` so serialization cannot reach them, in `notify` and anywhere else holding a credential
- Add a test asserting no credential material appears in `report --json`, `doctor --json`, or MCP responses, given a config that has them set
- While in there: decide whether `hasSecrets` should be a method on each config section rather than a function that has to be remembered when a section is added. The current shape is what let this drift, and it will drift again with the next credential-bearing section

## Notes

Surfaced by @gsaraiva2109 while reviewing where a Proxmox API token would live (#32). The Proxmox case is the same bug one step ahead: an inline token would bypass the permission gate exactly as notify credentials do today.

Worth fixing on its own rather than inside a feature PR, since it is a live gap and not a Proxmox concern.

Issue #64, opened 2026-08-22
Add Security policy


lenny-ts:
Add security policy and enable Private vulnerability reporting

Higangssh:
Good call, and sorry for the day of silence on it.

Private vulnerability reporting is on now, so the Security tab has a "Report a
vulnerability" button and there is a real channel for `SECURITY.md` to point at.
Doing it in that order was deliberate — a policy that names a reporting route
which does not exist yet is worse than no policy.

That half was mine because it is a repository setting. The document is yours if
you want it; assigning you on that basis, say the word and I will take it back.

What I would want in it:

**Supported versions: the latest release only.** This is a 0.x project shipping
every few weeks, there is no branch other than `main`, and nothing is backported.
Claiming a support window we do not honour is worse than admitting there is only
one. Worth revisiting at 1.0, not before.

**Response expectations, honestly.** One maintainer in one timezone. An
acknowledgement within a week is a promise that can be kept; 24 hours is not, and
this thread is evidence.

**What counts as a vulnerability here**, which is the part most policies skip and
the part that saves everyone time. In scope, roughly:

- credentials reaching somewhere they should not — command output, `--json`,
  report snapshots on disk, MCP responses. #61 is a live example
- the `serve` dashboard: auth bypass, token comparison, binding wider than asked
- the MCP surface letting an agent do more than the tool intends
- SSH and host key handling in `internal/remote`
- anything in `install` that runs code we did not intend to run

Out of scope, and worth saying so plainly: homebutler run as root can do what
root can do. That is the tool, not a finding.

Once it is in, the README should link it. Happy to review a draft at any size.

Higangssh:
Taking the document back, and I should not have assigned it to you without
asking in the first place — you raised the gap, which is not the same as
offering to write it.

Beyond that, on reflection this one is mine regardless. A security policy is not
documentation of how the code works, it is a commitment: which versions I will
fix, how fast I will answer, what I will treat as a vulnerability. Those are
promises only the person who has to keep them should be writing down. Handing
that to a contributor would mean you were drafting undertakings on my behalf.

The catch was the right one and the setting half is already done. #45, #38 and
#19 are still open and unassigned if you want the next thing.

Issue #45, opened 2026-08-21
Add a global --verbose flag so error output can lead with the fix


Higangssh:
Follow-up to #41. That issue raised two things; the remote half was a real bug and is
fixed in #44. This is the other half, which needs a design decision rather than a
message tweak.

## The problem

Reported in #41: when a command fails on a permission error, the raw Go error leads and
the actionable hint trails behind it.

```text
upgrading local... ✗ cannot backup current binary: rename /usr/local/bin/homebutler /usr/local/bin/homebutler.bak: permission denied

  ⚠️  Try: sudo homebutler upgrade
```

The hint is correct. The complaint is about ordering and noise — the only thing the user
needed was "rerun with sudo", and `rename /usr/local/bin/homebutler
/usr/local/bin/homebutler.bak: permission denied` tells them nothing they can act on.

The suggestion in #41 was to show just the fix and keep the raw error for verbose mode.
The second half is the blocker: there is no `--verbose`/`--debug` flag today, so dropping
the raw error would make it unrecoverable. I'd rather add the flag than lose the detail.

## Scope

This is not upgrade-specific. The same `raw error + hint` shape is used in nine places:

- `internal/remote/upgrade.go:107,119` — backup/write binary
- `internal/backup/backup.go:79,85` — create backup/compose dir
- `internal/backup/restore.go:127` — create bind mount dir
- `internal/install/install.go:39,45,63,683,692` — registry and app dirs
- `cmd/ports.go:31` — missing process names

Whatever we decide should apply to all of them, otherwise the output gets inconsistent.

## Why it isn't a one-line change

The messages are built inside `internal/*`, but a flag lives in `cmd/`. So this needs a
way for the internal packages to know the verbosity level. Options:

1. A package-level setting in `internal/util` that `cmd/root.go` sets before dispatch.
   Simplest, but it's global mutable state.
2. Thread a verbosity value through the call sites. Cleanest, but touches a lot of
   signatures for a presentation concern.
3. Keep returning the full error and strip the raw part at the presentation layer. Needs
   the two parts to be separable — a typed error carrying `{cause, hint}` rather than a
   pre-formatted string.

I lean toward (3): a small error type with a hint field, formatted at the point of
display. It fixes the ordering problem without any global state, and `--verbose` just
selects a different formatter. `util.IsPermissionError` already exists and is what gates
the hint today, so the detection side is done.

## Notes

[...]

Issue #38, opened 2026-08-17
Consolidate port parser tests into internal/ports and un-export the parsers


Higangssh:
## Description

`ss` and `lsof` parsing is now tested in two packages.

`internal/ports/ports_test.go` has nine tests against inline strings, covering both parsers, dedup, IPv6, header-only, empty input, and missing process info. `internal/inventory/portparse_fixture_test.go` (added in #35) tests the same two parsers against `testdata/` fixtures.

The split is my fault, not the contributor's. #29 asked for the fixtures under `internal/inventory/testdata/` on the mistaken belief that parsing lived in `internal/inventory`. It lives in `internal/ports`. The only thing genuinely pinned to `inventory` is the exposure classification, and once that moves in #37 there is nothing holding these tests outside the package that owns the code.

That split forced `parseLinuxOutput` / `parseDarwinOutput` to be exported purely so a test in another package could reach them. Both are now part of `internal/ports`' surface with no non-test caller.

## Scope

- Move `internal/inventory/testdata/{ss-linux,lsof-macos}.txt` to `internal/ports/testdata/`
- Move the fixture tests into `internal/ports` as in-package tests
- Revert `ParseLinuxOutput` / `ParseDarwinOutput` to `parseLinuxOutput` / `parseDarwinOutput`, and revert the call sites in `ports.go` and `ports_test.go`
- Fold the older inline-string tests into the fixture-driven ones where they overlap, keeping the edge cases that fixtures do not express (empty input, header-only, unprivileged output with no process column)
- Adopt #35's lookup-by-port-number style throughout. The existing tests index into `ports[0]` / `ports[2]`, which couples them to parser ordering
- Keep exposure assertions, calling `ports.IsPublicBind` from #37

## Notes

Worth doing as part of the JSON output schema work rather than on its own — that pass has to pin `PortInfo`'s serialized shape anyway, and these fixtures are where that gets asserted. #36 lands in the same files.

Depends on #37.

## Done when

- Parser tests live only in `internal/ports`
- The parsers are unexported again, with no non-test callers of the exported names left
- Coverage is not reduced relative to the union of the two current test files
- `go test ./...` passes

lenny-ts:
i'm taking this one

PR #65, opened 2026-08-22
refactor: single source of truth for inventory filters and --filter on show

### LINKED ISSUE #60: inventory show rejects --filter, and the supported-filter list has two sources of truth
## Description

Two follow-ups from reviewing #51, deliberately left out of that PR to keep it focused. Both live in the same two files and are worth doing together.

## `inventory show` advertises a flag it rejects

`--filter` is registered on `scan` only (`cmd/inventory.go`), while `show`'s `Short` says "Show current server inventory (same as scan)". So:

```
$ homebutler inventory show --filter exposed
Error: unknown flag: --filter
```

Both commands already share `runInventoryScan`, so this is either a one-line registration or a correction to the help text. Registering it on both is probably right — "same as scan" is the promise users read, and there is no reason the filter should not apply.

## The supported-filter list has two sources of truth

`RenderTreeFiltered` decides what it accepts with a `switch` (`internal/inventory/render.go:76`). `SupportedFilters()` returns a separate slice, and `IsSupportedFilter` derives from that slice. `cmd` validates against `IsSupportedFilter` before collection; `RenderTreeFiltered` validates again with the `switch`.

The double validation is correct — failing before collection is better than failing after. The problem is that the two lists can disagree. Add `"listening"` to `SupportedFilters()` and forget the `case`, and the pre-check passes, then the renderer returns:

```
unsupported filter "listening" (supported: exposed, listening)
```

An error that names the value it just rejected as supported.

Driving the `switch` from a map keyed by filter name, or making `IsSupportedFilter` the only gate, removes the trap. Either way the error string should be produced in one place rather than built independently in `cmd/inventory.go` and `render.go`.

## Notes

Neither is user-visible today with only one filter defined. Both become live the moment a second one is added, which is when nobody will be looking for them.

Follows #51, thanks @lenny-ts.

lenny-ts:
## What does this PR do?

Both follow-ups from reviewing #51, done together as one change:

**`inventory show` now accepts `--filter`.** The flag was registered on `scan` only
while `show`'s Short promises "same as scan". Both commands already share
`runInventoryScan`, so registration goes through a shared helper
(`addInventoryFilterFlag`) and the promise holds.

**One source of truth for supported filters.** The switch inside
`RenderTreeFiltered` and the `SupportedFilters()` slice were two lists that could
disagree — add a filter to the slice and forget the case, and the pre-check passes,
then the renderer returns `unsupported filter "listening" (supported: exposed,
listening)`, naming the value it just rejected as supported.

The two lists are gone. A single `filterRenderers` map keyed by filter name now
drives dispatch, validation, listing, and the error message:

- adding a filter is one map entry — everything else follows
- `SupportedFilters()` derives from the keys, sorted
- `IsSupportedFilter` is a map lookup
- the error string is built in exactly one place, new exported
  `UnsupportedFilterError(filter)`, used by both `cmd/inventory.go` and the renderer

The double validation itself stays: failing before collection is still the right
behavior — it just can't contradict itself anymore.

### Tests

- `TestRenderTreeFiltered_AgreesWithSupportedFilters` pins the invariant: everything
  `SupportedFilters()` lists must be accepted by the renderer, so drift can't return
- `TestNewInventoryShowCmdAcceptsFilter` — `show` parses `--filter exposed`
- `TestUnsupportedFilterError` — single error wording

README adds `homebutler inventory show --filter exposed` to the command block.

Closes #60 

## Checklist

- [x] README updated (if user-facing change)

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ Findings
 
 ```bash
 homebutler inventory scan
+homebutler inventory show --filter exposed
 homebutler inventory export --format mermaid
 homebutler --json inventory scan
 ```
--- a/cmd/inventory.go
+++ b/cmd/inventory.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/Higangssh/homebutler/internal/inventory"
 	"github.com/spf13/cobra"
@@ -24,6 +23,11 @@ func newInventoryCmd() *cobra.Command {
 	return invCmd
 }
 
+// addInventoryFilterFlag registers --filter on an inventory command.
+func addInventoryFilterFlag(cmd *cobra.Command, filter *string) {
+	cmd.Flags().StringVar(filter, "filter", "", "Filter inventory output (supported: exposed)")
+}
+
 func newInventoryScanCmd() *cobra.Command {
 	var filter string
 	cmd := &cobra.Command{
@@ -33,23 +37,26 @@ func newInventoryScanCmd() *cobra.Command {
 			return runInventoryScan(filter)
 		},
 	}
-	cmd.Flags().StringVar(&filter, "filter", "", "Filter inventory output (supported: exposed)")
+	addInventoryFilterFlag(cmd, &filter)
 	return cmd
 }
 
 func newInventoryShowCmd() *cobra.Command {
-	return &cobra.Command{
+	var filter string
+	cmd := &cobra.Command{
 		Use:   "show",
 		Short: "Show current server inventory (same as scan)",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInventoryScan("")
+			return runInventoryScan(filter)
 		},
 	}
+	addInventoryFilterFlag(cmd, &filter)
+	return cmd
 }
 
 func runInventoryScan(filter string) error {
 	if filter != "" && !inventory.IsSupportedFilter(filter) {
-		return fmt.Errorf("unsupported filter %q (supported: %s)", filter, strings.Join(inventory.SupportedFilters(), ", "))
+		return inventory.UnsupportedFilterError(filter)
 	}
 	if filter != "" && jsonOutput {
 		return fmt.Errorf("--filter is not supported with --json; JSON output is always the full inventory")
--- a/cmd/inventory.go
+++ b/cmd/inventory.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/Higangssh/homebutler/internal/inventory"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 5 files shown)


Higangssh:
Reviewed, and the code is clean — no correctness problems. The map dispatch is
initialised before any caller, it is read-only so the `--all` path stays
race-free, `show` closes over its own `filter`, and the double validation can no
longer contradict the renderer. Checked the three behaviours against a built
binary rather than just reading:

```
$ homebutler inventory show --filter bogus
error: unsupported filter "bogus" (supported: exposed)

$ homebutler --json inventory show --filter exposed
error: --filter is not supported with --json; JSON output is always the full inventory

$ homebutler inventory show --help
      --filter string   Filter inventory output (supported: exposed)
```

Deriving the flag usage string from `SupportedFilters()` was not in the issue.
That is the third time now you have closed the gap next to the one I asked
about.

**One thing to fix before I merge.** The description says `Closes #61`. #61 is
the `hasSecrets` issue — merging this as written would close a 1.0-blocking
security bug. It should be `Closes #60`.

Two more, neither blocking and both your call:

- `inventory_test.go:347` indexes `SupportedFilters()[0]` unconditionally. If the
  map is ever emptied during a refactor, that test panics instead of failing,
  which buries the real signal. Folding the check into the loop above it would
  do.
- `RenderTreeFiltered` lost its `Supported filters: "exposed" (only ports bound
  to public interfaces)` line, so someone reading the exported API can no longer
  see what `exposed` means — the semantics live only on the unexported renderer
  now. I would put that sentence on the `filterRenderers` entry rather than back
  in the function doc: a list in a doc comment is exactly the second source of
  truth this PR deletes.

The branch is also behind main. 0.21.0 through 0.21.2 went out today along with
a change to the release workflow, and the branch has to be current before it can
merge.

Minor, and the same as last time: there is a stray code fence at the end of the
description.

lenny-ts:
i think you can merge it now

Higangssh:
Merged. Thank you — genuinely, for the amount of attention you have put into
this project.

Three things landed today because of you, and none of them were the same kind
of work. #51 added the exposed filter and caught a README clause going stale
that I had not asked about. This one took both follow-ups and improved both
notes rather than just applying them: folding the index into the loop checks
every filter instead of the first, and putting `exposed`'s meaning on the map
entry rather than back in the doc comment kept the second source of truth this
PR deletes from quietly reappearing. That distinction was the whole point and
you did not need it spelled out twice.

And #64 was the one with the longest reach. Private vulnerability reporting is
on, `SECURITY.md` is written, and someone who finds a problem in homebutler now
has somewhere to send it that is not a public issue. That gap had been open
since the first release and it took someone else noticing.

Three of the four things I said were fair game are still open — #45, #38 and
#19. No pressure on any of them, and none is more welcome than the next. If
none of those appeal, opening an issue for something you hit while using it is
just as useful; #64 is the evidence.

Issue #32, opened 2026-07-25
feat: add Proxmox VE cluster integration


gsaraiva2109:
## Problem

HomeButler already gives a useful single-server and SSH multi-server view, but a Proxmox VE homelab is operated as a cluster: nodes, LXC containers, VMs, storage, and tasks need to be visible together. Today this requires switching to the Proxmox UI or shell for both monitoring and provisioning.

## Proposed solution

Add an optional Proxmox VE integration that uses the official API and fits the existing CLI, JSON, web dashboard, and MCP interfaces. Deliver it in independently useful phases:

1. **Read-only cluster visibility**
   - Configure one or more Proxmox API endpoints using API tokens.
   - Show cluster, node, storage, LXC, and VM health and resource usage.
   - Surface guest state, host node, uptime, CPU, memory, disk, and recent tasks.
   - Support JSON output and a concise cluster summary in the dashboard and MCP tools.

2. **Explicit guest actions**
   - Start, stop, reboot, and inspect task status for an LXC or VM.
   - Require an explicit target and confirmation for disruptive actions.

3. **Community-script provisioning**
   - Let a user select an entry from the official `community-scripts/ProxmoxVE` catalog, choose the target node and resources, review the exact command, then explicitly run it.
   - Track the Proxmox task and return the created guest ID, address when available, and failure output.
   - Do not run arbitrary remote scripts automatically or silently.

Example direction:

```text
homebutler proxmox status --json
homebutler proxmox guest list
homebutler proxmox lxc create --script uptime-kuma --node pve1
homebutler proxmox task show <task-id>
```

## Security and design constraints

- Use scoped API tokens and document minimum permissions.
- Keep the first phase read-only by default.
- Verify TLS by default; any self-signed-certificate exception must be explicit and clearly warned.
- Keep API credentials out of command output, reports, and MCP responses.
- Preserve the project’s single-binary and low-dependency approach.
- Reuse existing multi-server output, JSON conventions, alerts, and dashboard patterns where possible.

## Alternatives considered

SSHing into every Proxmox node could expose some data, but it is less reliable for cluster-wide state and makes task tracking and least-privilege access harder. The Proxmox API is a better foundation for a cluster integration.

## Why this matters

This would make HomeButler a practical monitoring and control layer for an entire Proxmox cluster, while preserving the safe, structured interface that makes it useful from scripts and AI tools. The community-script flow would also make common LXC deployments discoverable and repeatable without replacing the Proxmox UI.

References:

[...]

Issue #62, opened 2026-08-22
Decide whether homebutler executes externally-authored provisioning scripts (Proxmox Phase 3)


Higangssh:
## Description

Split out of #32 as promised there, so a monitoring feature is not carrying a security decision along with it.

The Proxmox proposal included a third phase: provisioning guests through the Proxmox VE Community Scripts catalog. It is a genuinely useful idea and it is also unlike anything homebutler does today, which is why it needs its own discussion rather than a line in a Phase 1 PR.

## Why it is different from `install`

`homebutler install` looks superficially similar — pick a name, get a running thing — but the trust model is not comparable.

Today `install` renders a compose template that lives in this repository (`internal/install/install.go`), hands it to Docker, and whatever runs stays inside a container. The catalog is curated in-tree, so what executes is whatever was reviewed when the entry was merged.

Phase 3 would fetch a shell script from an external catalog at runtime and execute it as root on the hypervisor. Three things change at once: the code is not reviewed by us, it is fetched at run time rather than pinned at build time, and it runs as root outside any container. It would also give homebutler its first code path that executes arbitrary shell — something deliberately avoided in #19, where `docker exec` was rejected on the same grounds.

## What needs deciding

- Whether homebutler runs externally-authored code at all, or stops at generating and showing the command for a human to run
- If it does: pinning. A catalog reference by commit or checksum, verified before execution, versus fetching whatever is at a URL today
- What the MCP story is. `riskDestructive` is the obvious classification, but an agent invoking root-level provisioning deserves more than a label — that likely needs the permission model that does not exist yet
- Whether the answer differs for a script the user names explicitly versus one homebutler selects from a catalog

## Notes

No decision implied by opening this. It is here so #32 can land Phase 1 and 2 without settling it, and so the question is on the record rather than in a comment thread.

Raised by @gsaraiva2109 in #32.

Higangssh:
**Decided: homebutler does not execute externally-authored code. Phase 3 resolves and prints the command; a person runs it.**

The reasoning is mostly that I already decided this once. #19 rejected `docker exec` because arbitrary execution contradicts what the README promises — "readable summaries and JSON instead of asking you to trust a black-box shell session." Phase 3 is the same shape and further along it:

| | `docker exec` (rejected) | Phase 3 execution |
| --- | --- | --- |
| Where the code comes from | typed by the operator | an external catalog |
| When it is reviewed | — | never; fetched at run time |
[...]

Issue #54, opened 2026-08-22
Close the MCP coverage gap before 1.0 freezes the tool surface


Higangssh:
Twenty-three MCP tools cover `status`, `docker`, `ports`, `network`, `wake`,
`alerts`, `inventory`, `report`, `doctor`, `backup`, and `install`. Four
commands have no tool at all, and one of them is the one the README leads with.

## The gap

`watch` is missing entirely. The README asks "Why did this service restart at
3 AM?" and `watch` is what answers it, so the question an agent is most likely
to ask is the one it cannot. `report` answers what changed and `watch` answers
what broke; only half of that reaches MCP today.

`processes` is missing too, and "what is eating the CPU" is near the top of
what anyone asks a server.

## Tools to add

| tool | risk | remote | arguments |
| --- | --- | --- | --- |
| `watch_check` | write | yes | `server` |
| `watch_history` | read | yes | `server`, `limit`, `container`, `include_logs` |
| `watch_list` | read | yes | `server` |
| `processes` | read | yes | `server`, `limit`, `sort_by` |
| `config_validate` | read | no | `strict` |

Three decisions worth settling before any of this is written, since 1.0 freezes
the tool surface:

**`watch_check` is `riskWrite`, not `riskRead`.** It writes `states` and saves
incidents under the watch directory. It reads like a query but it is not one.
`report` is already classified this way for the same reason — it saves a
snapshot.

**`watch_history` excludes logs by default.** Every `Incident` carries
`pre_logs` and `post_logs`, which is 100 lines of container output each. Thirty
incidents of that will bury an agent's context in log text it did not ask for.
`limit` defaults to 10 and `include_logs` defaults to false, so the caller opts
in to the expensive shape.

**`config_validate` is local-only.** It answers "is the config this MCP server
is running on valid", which is a question about this machine. Routing it to a
remote would silently answer a different question.

## Deliberately excluded

`trust` must not be exposed. It accepts SSH host keys on first use, and an
agent auto-accepting TOFU is exactly the boundary the project promises not to
cross. Worth stating in the README rather than leaving as an omission.

`upgrade` replaces the running binary. `serve`, `init`, `watch start`, and
`watch tui` are daemons or interactive; neither fits a stdio request/response
tool.

`deploy` is deferred past 1.0. Remote install is the highest-risk surface here
and should not be frozen in the first stable release.

## Shape of the work

Each tool touches four places, with `doctor` as the reference implementation:

[...]

PR #78, opened 2026-08-24
feat: add read-only Proxmox visibility

gsaraiva2109:
## feat: add read-only Proxmox visibility

Part of #32

## Why

HomeButler could inspect local and SSH-managed hosts, but it had no API-native way to inspect a Proxmox VE cluster. Treating Proxmox as another `servers:` entry would have mixed SSH deployment settings with API credentials, certificate trust, and cluster resources.

This keeps Proxmox as a separate named endpoint. A Proxmox endpoint is the target, so the commands and MCP tools do not join the SSH `--server` or `--all` fan-out paths.

## Design

The new `proxmox:` config section supports one or more named endpoints. It requires a host, token ID, and one token source. `token_file` is preferred; inline `token` remains protected by the existing generic secret handling and is excluded from JSON serialization.

The client uses only the Go standard library. TLS verification stays enabled by default and accepts trust in this order:

1. SHA-256 certificate fingerprint pinning
2. a configured CA file
3. `insecure: true` as an explicit fallback

The client exposes the read-only version, cluster status, resources, node status, and recent tasks endpoints. It preserves Proxmox data semantics that matter to callers: unified QEMU and LXC guests, optional metrics, CPU fractions, large byte counters, tags, templates, and boolean wire variants.

## Read-only interfaces

CLI commands:

- `homebutler proxmox status [--endpoint <name>] [--json]`
- `homebutler proxmox guests [--endpoint <name>] [--node <n>] [--status running|stopped] [--json]`
- `homebutler proxmox node <name> [--endpoint <name>] [--json]`
- `homebutler proxmox tasks [--endpoint <name>] [--node <n>] [--limit 50] [--json]`

When exactly one endpoint is configured, `--endpoint` is optional. Multiple endpoints require it. Human status output includes cluster quorum and node health; JSON returns the typed API data.

MCP adds four `riskRead` tools targeted only at named Proxmox endpoints:

- `proxmox_status`
- `proxmox_guests`
- `proxmox_node`
- `proxmox_tasks`

The MCP tools follow the same endpoint-selection rules and return typed data without configuration or token values.

## Tests and fixtures

The client and both interfaces are backed by sanitized recorded-response fixtures. Coverage includes standalone and clustered status, resources with missing and zero metrics, tags, templates, locks, large counters, task queries, authentication headers, TLS fingerprint and CA verification, explicit insecure mode, permission errors, and empty ACL-limited resource responses.

The CLI and MCP tests exercise HTTPS fixture servers, configured token resolution, endpoint selection, filtering, request paths, query parameters, and tool capability registration.

## Deliberately not included

- Guest actions are Phase 2.
- The web dashboard is a separate deferred workstream until the data shape settles.
- Proxmox endpoints do not participate in SSH `--server` or `--all` fan-out.
- Community Scripts provisioning is Phase 3.

## Verification

- `go vet ./...`
- `golangci-lint run`
- `go build ./...`
- `go test ./cmd ./internal/config ./internal/mcp ./internal/proxmox`
- `go test ./... -skip '^TestRemoveHostKeys_NonexistentFile$'`
- read-only live `proxmox status --json` smoke test against a configured endpoint

The skipped upstream test can rewrite the developer's real SSH known-hosts file. All remaining packages pass.


cmd/proxmox.go
```diff
@@ -0,0 +1,252 @@
...
+			if err != nil {
+				return err
+			}
+			nodes := []string{node}
+			if node == "" {
+				resources, err := client.Resources(context.Background())
+				if err != nil {
+					return fmt.Errorf("get Proxmox nodes: %w", err)
+				}
+				nodes = make([]string, len(resources.Nodes))
```
Higangssh, line 157: When the token's ACLs hide every node, `resources.Nodes` is empty, `nodes` is length 0, the loop below never runs, and the command prints the `Tasks` heading and nothing else. That reads as "no recent tasks" when nothing was queried.

This is the shape #46 fixed for `watch check` — it counted registered targets while inspecting only docker ones, so a systemd-only watch list printed "Checked 1 container(s). No restarts detected." having looked at nothing.

Worth saying which nodes were queried, or reporting that none were visible. The `cluster-resources-empty-acl.json` fixture already exists; nothing exercises this path through it.

cmd/proxmox.go
```diff
@@ -0,0 +1,252 @@
...
+				for i, resourceNode := range resources.Nodes {
+					nodes[i] = resourceNode.Name
+				}
+			}
+			tasks := make([]proxmox.Task, 0)
+			for _, taskNode := range nodes {
+				entries, err := client.TasksLimit(context.Background(), taskNode, limit)
+				if err != nil {
+					return fmt.Errorf("get Proxmox tasks for node %q: %w", taskNode, err)
```
Higangssh: One unreachable node loses the tasks from every other node. On a three-node cluster with one down, this returns an error and no tasks at all.

#74 settled the convention for this two days ago: collect best-effort, record what failed. `internal/inventory/inventory.go:81-96` is the pattern — `Warnings` carries the human-readable half and `Failed` the machine-readable one, so a reader can tell an empty result from a missing one.

Same applies to `DefaultView` (`internal/proxmox/client.go:212`): version, cluster status and resources run in sequence and the first error discards the two that succeeded. A token that can read `/version` but not `/cluster/resources` currently shows nothing.

internal/mcp/proxmox.go
```diff
@@ -0,0 +1,81 @@
...
+	case "proxmox_node":
+		return client.NodeStatus(ctx, stringArg(args, "node"))
+	case "proxmox_tasks":
+		return client.Tasks(ctx, stringArg(args, "node"))
+	default:
+		return nil, fmt.Errorf("unknown tool: %s", name)
+	}
+}
+
+func selectProxmoxEndpoint(c *config.Config, name string) (*config.ProxmoxConfig, error) {
```
Higangssh: This is the second copy of the rule — `cmd/proxmox.go:194` has the same four branches. They already differ: the CLI names the available endpoints in the not-found and multiple-endpoint errors, this one does not.

Two independent copies of one selection rule is what #47 consolidated for `isPublicBind` and #65 for the filter list. Nothing breaks today; it breaks the first time the rule changes and only one copy moves.

`config` is imported by both, so `config.SelectProxmox` or similar would leave one gate and one error string.

Higangssh (commented):
Read the whole thing. The design is what we agreed in #32, and the parts I was most likely to argue with are the parts you got right: the fingerprint path disables chain and hostname verification deliberately and pins in constant time, the trust order is early-return so a safer setting always wins, and the `secret:"true"` tag means #75's permission gate and JSON exclusion picked up the inline token without either of us wiring it. The registry entries are the shape I described. Standard library only. Fixtures and docs included.

CI is green — lint, tests and all four build targets. It needed workflow approval first, which is why the checks were blank until now.

**One change needed before merge, unrelated to the code.** The body says `Closes #32`. Please make it `Part of #32`. Phase 2 and Phase 3 have not started, #62 is still open on the Phase 3 question, and #32 is where both hang. Merging as-is closes the umbrella out from under them.

Scope question answered in #32: the dashboard stays out, Phase 1 is complete without it, tracked as #79.

The three inline comments are what I would like fixed here. Two of them are conventions this repository settled in the last two weeks, so they are less about Proxmox than about not drifting from #46 and #74.

Everything below is follow-up material, not blocking. Several are mine to fix rather than yours; say the word and I will open them.

- `insecure: true` is silent. We agreed in #32 that it should be "last and loud", and it is last but not loud — no runtime warning, no `config validate` finding, no `doctor` check. `serve` warns on a wildcard bind (#63) and this is the same class.
- `config validate` checks that `token_file` exists but never looks at `ca_file`, and does not validate the fingerprint format. Both fail later at client construction, so `config validate` passes and `proxmox status` does not.
- `TimeoutDuration` (`internal/config/config.go:158`) says validation reports malformed values before a client is constructed. `Load` never calls `Validate` — that is a separate command a user has to run — so `timeout: "abc"` silently becomes 10s. Either the comment or the behaviour should move.
- `--json` indentation splits: `status` indents, `guests`, `node` and `tasks` do not. `output()` indents everywhere else.
- `io.LimitReader(resp.Body, 1<<20)` (`internal/proxmox/client.go:246`) does not detect truncation, so a cluster past 1MB of resources gets "unexpected end of JSON input". Reading one byte past the cap and failing explicitly would name the real problem.
- `context.Background()` rather than `cmd.Context()`, so Ctrl+C does not cancel. The client timeout bounds it, so this is cosmetic.
- MCP `proxmox_tasks` has no limit parameter while the CLI has `--limit`. Worth noting only because #54 is about parity.

Thanks for asking about the dashboard instead of quietly dropping it. That is the part that made this easy to review.

gsaraiva2109:
Implemented in 5a9a1ea.

- Centralized Proxmox endpoint selection in `config.SelectProxmox`, so CLI and MCP now share one rule and error message.
- Made `proxmox tasks` collect per-node results best-effort, exposing queried nodes plus warnings and failed nodes; it now explicitly reports when ACLs expose no nodes.
- Made `DefaultView` retain successful version, cluster, or resource responses and report partial failures through `warnings` and `failed`.

The PR body already says `Part of #32`, so it needs no change.

Verified with `go vet ./...`, `go test ./... -skip '^TestRemoveHostKeys_NonexistentFile$'`, `golangci-lint run`, and `go build ./...`.

Higangssh:
All four are in, and the partial-failure handling reads better than what I asked for. `Version: unavailable` per section with the warnings printed underneath means a fully failed connection can no longer pass for a healthy cluster of zero, which was the part I was most worried about. CI is green.

Two naming details before this lands, both small, and both because this PR is the first place `Warnings`/`Failed` appears outside `inventory` — whatever it settles on here is what #79 and everything after will copy.

**The JSON field name splits.** `inventory` serializes it as `failed_collectors` (`internal/inventory/inventory.go:25`); `DefaultView` and `proxmoxTasksView` use `failed`. Two keys for one concept undoes some of what #74 was for, which was giving a reader a machine-readable field instead of prefix-matching warning strings. Please use `failed_collectors` in both.

**The collector names are string literals.** `"version"`, `"cluster"` and `"resources"` appear in `client.go` and again in three `CollectorFailed(...)` calls in `cmd/proxmox.go`. `inventory` declares `CollectorDocker` and `CollectorPorts` as constants for this. Worth copying: `CollectorFailed("resource")` returns false silently, so a typo renders a failed section as available with zeros — the failure mode this PR just fixed, reintroduced by a misspelling.

`proxmoxTasksView.Failed` holding node names rather than collector names is right as it is; nodes are the unit of collection there.

Nothing else from my side. Once those two are in I will merge.

gsaraiva2109:
Implemented in 7f16917.

`DefaultView` and `proxmoxTasksView` now serialize failures as `failed_collectors`. The task view still records node names, as intended.

Added `CollectorVersion`, `CollectorCluster`, and `CollectorResources` constants and used them in the client and status renderer. Focused tests cover both JSON field names.

Verified with `go test ./cmd ./internal/proxmox`, `go vet ./...`, `golangci-lint run`, and `go build ./...`.

Thanks, this keeps the collection contract consistent before #79 adopts it.

Higangssh:
Merged as 3edd0ee. Thank you — this was a pleasure to review.

Three things I want to name, because they are the reason it went in fast. You checked what the capability registry field actually did before proposing a replacement, rather than assuming. You asked about the dashboard instead of quietly dropping it, which turned a scope change into a decision with a record. And you locked the `failed_collectors` name with a test that also asserts the old key is gone, which is more than I asked for and the right call for a field that 1.0 will freeze.

The TLS work is the part I was most prepared to argue with and had nothing to say about.

Phase 2 whenever you are ready. #62 is the one worth settling before Phase 3 gets written, and your operating experience is most of what that thread is missing.

PR #82, opened 2026-08-24
feat: answer container diagnostics from the host with docker top and inspect

### LINKED ISSUE #19: Add container diagnostics without exec
## Description

Make the common container troubleshooting questions answerable through
homebutler, without adding a general command-execution primitive.

This issue originally asked for `homebutler docker exec <container> -- <cmd>`.
That scope is withdrawn — see the comment below for the reasoning. The short
version is that the two examples it was motivated by are both answerable from
the host, and arbitrary exec would contradict what the rest of the tool
promises.

## Expected behavior

```bash
$ homebutler docker top app-api
PID     USER    COMMAND
1       root    nginx: master process nginx -g daemon off;
31      nginx   nginx: worker process
```

```bash
$ homebutler docker inspect app-api
📦 app-api
   Image     nginx:1.27-alpine
   State     running (up 4d)
   Restart   unless-stopped · 0 restarts
   Ports     0.0.0.0:8080 → 80/tcp
   Mounts    /srv/app/data → /data (rw)
   Networks  bridge (172.17.0.4)
   Health    healthy
```

```bash
homebutler docker top app-api --json
homebutler docker inspect app-api --json
```

## Scope

- Add `docker top <container>` — processes running inside a container, read
  from the host via `docker top`. No exec, no TTY.
- Add `docker inspect <container>` — a readable summary of image, state,
  restart policy, ports, mounts, networks, and health, from `docker inspect`.
- Both support `--json`, and both are read-only.
- Expose both through MCP as `riskRead` capabilities.

## Out of scope

- **Arbitrary command execution.** Not in this issue and not behind a flag.
  If a diagnostic genuinely cannot be served from the host, open a separate
  issue for that specific need.
- **Environment variables.** `docker inspect` exposes them, and container env
  routinely holds database passwords and API tokens. Do not include them in
  either output. If they are ever added, it must be keys only, values
  redacted, and it needs its own discussion before going anywhere near MCP.

## Implementation hints

- Follow the shape of `Logs` and `Stats` in `internal/docker/docker.go`:
  validate the name with `isValidName`, shell out through `util.DockerCmd`,
  return a typed struct with json tags, keep parsing in a pure function.
- Render with a `format.*` function in `internal/format/human.go`, wired
  through the `output()` type switch in `cmd/helpers.go`.
- Add the subcommands to `cmd/docker.go` alongside `logs` and `stats`.
- `docker top` passes its arguments to `ps` on the host, so column layout
  differs between platforms. Parse defensively and commit a captured sample
  under `testdata/` rather than asserting against a live daemon — same
  reasoning as #29.
- `docker inspect` returns a JSON array; decode into a minimal struct holding
  only the fields above rather than mirroring the whole schema.

## Done when

- `homebutler docker top <container>` and `homebutler docker inspect <container>`
  work against a running container, in human and `--json` form
- A missing container, a stopped container, and a docker daemon that is not
  running each produce a clear error
- Neither command emits environment variable values
- Both are reachable over MCP as read-only capabilities
- Parsing is covered by fixture tests
- `go test ./...` passes

lenny-ts:
Closes #19

The two questions behind most container troubleshooting — *what is running in
there* and *how is that container configured* — are now answerable without
giving up the no-exec promise:

    $ homebutler docker top app-api
    PID      USER         COMMAND
    1        root         nginx: master process nginx -g daemon off;
    31       nginx        nginx: worker process

    $ homebutler docker inspect app-api
    📦 app-api
       Image     nginx:1.27-alpine
       State     running (up 4d)
       Restart   unless-stopped · 0 restarts
       Ports     0.0.0.0:8080 → 80/tcp
       Mounts    /srv/app/data → /data (rw)
       Networks  bridge (172.17.0.4)
       Health    healthy

Both support `--json`, both are read-only, and both are reachable over MCP as
`riskRead` capabilities pointed at local or server targets.

## Decisions a reviewer might ask about

**Env values are excluded by construction, not by filtering.** The decode
target simply has no field for `Config.Env`, so there is no code path that
could leak one. The fixture is derived from a real `docker inspect` with a
planted password in env, and a test asserts it never surfaces in either output
form.

**`docker top` parsing assumes only what every layout shares.** The output is
host `ps` output, so columns differ per platform; the parser reads user from
the first column, finds PID by name, and treats everything from the last header
column onward as the command. Anything else is skipped rather than
misindexed. Fixtures are verbatim captures — Linux `ps -ef` from a live daemon,
plus a macOS `ps aux` variant — instead of assertions against whatever `ps`
the test host runs.

**Container names can no longer look like CLI flags.** While wiring these two
commands, a flag-shaped name arriving over MCP (`--tls`, `-ldebug`) was found
reaching the docker binary as argv rather than as an argument. Existing checks
only blocked the dangerous forms by accident, because the character set
excludes `=`, `:` and `/`. `isValidName` now rejects a leading dash explicitly,
mirroring docker's own grammar; restart, stop and logs validate through the
same check.

**Errors carry the daemon's reason, not just an exit code.** A missing
container reports `no such object: <name>`, a stopped one reports that it is
not running, because `CombinedOutput` already held those words.

**Uptime takes an injected clock** (`parseDockerInspect(out, now)`), so the
fixture pins `up 4d` deterministically instead of drifting as the sample ages.

Demo mode mirrors the existing fleet (nginx, postgres, redis…) keyed by
container name with a fallback entrypoint row, matching how `demoLogs` behaves,
so the registry-walking test exercises both new arms through the schema.


internal/docker/docker.go
```diff
@@ -85,6 +89,271 @@ func Logs(name string, lines string) (*LogsResult, error) {
...
+					pidCol = i
+					break
+				}
+			}
+			if pidCol < 0 || cmdCol-1 <= pidCol {
+				return processes
+			}
+			continue
+		}
+		// A row shorter than the header has no reliable command tail; skip it.
```
Higangssh: The choice is right — misindexing a row would be worse than dropping it, and `TestParseDockerTopDefensive` pins it. What is missing is telling the reader it happened. `homebutler docker top app-api` showing two of three processes reads as "this container has two processes".

This repository has called that shape a bug twice recently: #46, where `watch check` counted targets it never inspected, and #78, where `proxmox tasks` printed an empty list for a node set it could not see. Both were fixed by saying so, not by parsing harder.

A `Skipped int` on `TopResult` with `omitempty`, plus a line in `format.DockerTop`, keeps the parser exactly as conservative as it is now and stops it being quiet. The same applies to the `pidCol < 0` return above: an unrecognised ps layout currently renders identically to a container with no processes.

internal/docker/docker.go
```diff
@@ -150,15 +419,18 @@ func shortenDuration(s string) string {
 	return s
 }
 
-// isValidName prevents command injection by allowing only safe characters.
+// isValidName prevents command injection by allowing only safe characters,
+// and blocks a leading dash so a name can never be parsed as a docker CLI
+// flag rather than an argument (CWE-88). The first character rule mirrors
```
Higangssh: Small wording thing — the comment claims a little more than the line enforces. `[a-zA-Z0-9][a-zA-Z0-9_.-]*` also excludes a leading `_` and `.`, and both still pass here; only the leading dash is rejected.

Harmless in practice, since docker rejects those names itself and the dash is the only character that creates the flag ambiguity this is about. Worth trimming the claim to what the check actually does, so the next reader does not assume the full grammar is covered. Take it or leave it.

Higangssh (commented):
This is a nice piece of work, and several things I would normally have to ask for are already here.

Excluding env by construction rather than by filtering is better than what the issue asked for — a filter is something a future edit can forget, a field that does not exist in the decode target is not. Planting a real-looking secret in the fixture and asserting on it is what makes that hold. The injected clock so `up 4d` does not drift, and verbatim ps captures from two platforms instead of assertions against whatever ps the test host runs, are both the kind of thing that only shows up when someone has thought about the tests aging.

Finding the flag-shaped-name path while wiring this up, and then fixing it for `restart`, `stop` and `logs` rather than only the two new commands, is more than the issue asked for and the right instinct.

CI is green, and `gofmt`, `go vet` and the package tests pass locally too.

**One thing before merge.** The PR title is `Feat/docker top inspect`, taken from the branch name. This repository squash-merges with the PR title as the commit subject, so that is what would land in `git log`. Your first commit already has the subject I would want:

```
feat: answer container diagnostics from the host with docker top and inspect
```

Could you set that as the PR title?

**One thing I would like fixed here** — the inline comment on `parseDockerTop`. The skipping is correct, but it is silent, and this repository has twice decided a silent skip is a bug. Small addition, and you are already in that file.

The second inline comment is a wording nit on the `isValidName` doc comment; ignore it if you disagree.

One observation that is not yours to fix: the leading-dash rejection only guards the local path. `executeRemote` forwards the raw name into the remote homebutler's argv, so a flag-shaped name becomes a cobra error on the far side and never reaches `isValidName`. Not exploitable — the far side is homebutler, not docker — and it predates this PR for `docker_logs`, `docker_restart` and `docker_stop` as well. I will open a follow-up so both paths report the same thing.

Thanks for picking this one up.

Higangssh:
Opened the remote-path follow-up as #83, so it is tracked rather than just mentioned. Nothing there is yours — it covers `docker_restart`, `docker_stop` and `docker_logs` as they already are on main, and `docker_top` and `docker_inspect` inherit it from the same routing.

lenny-ts:
Both addressed in 91b576c.

On the silent skip: `TopResult.Skipped` now counts both shapes of drop — a row
shorter than its header, and every row after an unrecognised ps header — so an
unparseable layout no longer renders identically to a container with no
processes. `format.DockerTop` prints "N of M output rows could not be parsed"
under the table, and the JSON carries `skipped` only when it is non-zero. The
parser is exactly as conservative as before; it just reports what it dropped.

On the doc comment: took it. The claim now says only what the check enforces —
a leading `_` or `.` still passes here, and the dash is the only leading
character that creates flag ambiguity.

Thanks for opening #83 for the remote path — agreed that nothing there blocks
this PR, and one consistent rejection across both paths is better fixed once
than patched here twice.

Higangssh:
Merged as 930be68, and #19 closes with it. Fifth one — thank you.

Two things I want to record, because they are why this took one round.

Excluding env by construction was better than the issue asked for. #19 said "do not include them", which describes a filter, and a filter is something a later edit forgets. A decode target with no field for `Config.Env` cannot forget. Planting a real-looking secret in the fixture and asserting on it is what turns that from an intention into a property.

And you rewrote `parseDockerTop` rather than bolting a counter onto it. The old version carried its state in a `cmdCol == 0 && pidCol == 0` sentinel; collecting the rows first and treating `rows[0]` as the header made the skip count fall out naturally instead of needing to be threaded through. Fixing the `pidCol < 0` path too, which I mentioned almost in passing, is the difference between addressing a comment and addressing the problem behind it.

You also improved the doc-comment change I suggested. I would have deleted the claim; saying that docker's grammar is stricter and that the dash is the only leading character creating flag ambiguity is more useful than saying less.

#83 is open whenever you want it, though it is nobody's in particular. Same underlying decision you named in your comment: one gate rather than two.

PR #88, opened 2026-08-26
fix: refuse flag-shaped container arguments before routing to a remote host

### LINKED ISSUE #83: A flag-shaped container name skips validation on the remote path, so docker_restart can return help text
## Description

`docker.Restart`, `docker.Stop` and `docker.Logs` all validate the container name with `isValidName` before shelling out, and #82 adds the same check to `Top` and `Inspect`. That check only guards the local execution path.

`executeRemote` (`internal/mcp/server.go:483`) forwards the name straight through:

```go
case "docker_restart":
    remoteArgs = []string{"docker", "restart", stringArg(args, "name"), "--json"}
```

`remote.Run` shell-quotes every argument with `util.ShellQuoteArgs`, so **this is not shell injection** — the value arrives on the remote host as a single argv element. But that argv element is then parsed by the remote homebutler's own cobra command, and cobra reads a leading dash as a flag. `isValidName` runs after cobra has already decided, so it never sees the value.

## What it looks like

An agent calling `docker_restart` with `server: "media"` and `name: "--help"` gets the remote command:

```
homebutler docker restart --help --json
```

cobra prints the help text for `docker restart` and exits zero, so `remote.Run` succeeds and the MCP tool returns homebutler's help as if it were a restart result. No error, no indication that nothing was restarted.

Locally the same call returns `invalid container name: --help`. Same tool, same argument, two different outcomes depending on whether `server` was set.

## Not a vulnerability

Worth stating plainly so nobody reads this as one. The far side is homebutler, not docker: a flag-shaped name reaches a cobra parser, not a docker CLI, and the arguments are shell-quoted before they leave. The worst case is a confusing result, not code execution or a container being acted on unexpectedly. It is a correctness and consistency bug.

## Scope

Five tools forward a name this way: `docker_restart`, `docker_stop`, `docker_logs`, and since #82 merged, `docker_top` and `docker_inspect`.

`docker_logs` also forwards `lines` raw (`internal/mcp/server.go:496-500`), which has the same shape — a `lines` value of `--json` becomes a flag rather than a bad number.

The fix is deciding where validation belongs. Validating in `executeRemote` means the rule lives in two places, which is what #47 and #65 were both about. Validating once in `executeTool`, before the local-or-remote decision, would leave one gate. `isValidName` is unexported in `internal/docker`, so that choice also decides whether it gets exported or moves.

## Done when

- A flag-shaped name is refused with the same message whether or not `server` is set
- The rule has one home rather than one per path
- `docker_logs` rejects a non-numeric `lines` before sending it
- A test covers the remote path, not only the local one

## Notes

Found while reviewing #82, which added the leading-dash rejection to `isValidName` after noticing flag-shaped names reaching the docker binary locally. That fix is correct and this is the half it does not reach. Not #82's to fix — the gap predates it for the three existing tools.

lenny-ts:
A name like `--help` sent to `docker_restart` with `server` set travelled over
as a single shell-quoted argv element and was parsed by the remote homebutler's
own cobra command, which read it as a flag: the help text came back with exit
zero and looked like a restart result. Locally the same call was refused, so
the same tool answered differently depending on whether `server` was set (#83).

Validation now runs once in `executeTool`, before the local-or-remote decision.
The rules themselves stay in `internal/docker` — `isValidName` becomes exported
`ValidName`, and the digits-only check that lived inline in `Logs` becomes
`ValidLines` — so there is one copy of each rule and the gate only applies it.
Rejection messages come from the gate, so remote answers match the local ones
word for word rather than by coincidence.

Two decisions worth calling out:

- The checks inside `Restart`, `Stop`, `Logs`, `Top` and `Inspect` stay. They
  are not part of the two-paths problem: `homebutler docker restart -- --help`
  and the alert remediation reach these functions directly, and that is where
  the name meets a real argv.
- `ValidLines("")` is false where the old inline loop passed an empty string
  through vacuously. Every caller defaults to `"50"` before validating, so
  nothing observable changes; the helper is honest on its own now. On the
  remote path this closes the same shape for `lines`: `--json` used to become
  a flag of the remote command, it is now refused locally with the message the
  local path already gave.

Missing names get the same treatment: they were previously forwarded as an
empty argument, and both paths now answer `missing required parameter: name`.

Tests exercise the remote path without opening a connection: the five tools
are pointed at a configured server and the assertion is that the rejection
carries the validation message rather than a dial error or a success.
`TestForwardedContainerNamesHaveAGate` reads `executeRemote`'s switch the way
`TestRegistryTargetsMatchRemoteArgvMapping` does, so a future tool forwarding
`name` cannot skip the gate silently.

Closes #83.

### DIFF
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -254,14 +254,14 @@ func validateContainerArgs(tool string, args map[string]any) error {
 }
 
 func (s *Server) executeTool(name string, args map[string]any) (any, error) {
-	if s.demo {
-		return s.executeDemoTool(name, args)
-	}
-
 	// Before any routing: whichever machine ends up answering, the argument
 	// rules are the same and were already checked.
 	if err := validateContainerArgs(name, args); err != nil {
 		return nil, err
+	}
+
+	if s.demo {
+		return s.executeDemoTool(name, args)
 	}
 
 	if cap, ok := capabilityFor(name); ok && cap.supports(targetProxmox) {
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -206,9 +206,61 @@ func (s *Server) handleToolCall(req *jsonRPCRequest) {
 	})
 }
 
+// containerArgTools lists the tools whose container name travels into a
+// command line, here or on a remote host, and names the second value they
+// forward alongside it where one exists. executeTool validates each of them
+// before choosing a path, because the paths fail differently otherwise: the
+// local switch lands in the docker package, which rejects a bad name with
+// these exact words, while the remote path builds an argv for another
+// homebutler whose own command parser reads a leading dash as a flag — the
+// help text then comes back through remote.Run as if a restart had happened.
+var containerArgTools = map[string]string{
+	"docker_restart": "",
+	"docker_stop":    "",
+	"docker_logs":    "lines",
+	"docker_top":     "",
+	"docker_inspect": "",
+}
+
+// validateContainerArgs is the one gate for forwarded container arguments,
+// applied before the local-or-remote decision so neither path can grow its
+// own dialect of the rule. The error messages match what the docker package
+// returns on the local path, so a caller cannot tell which machine would have
+// answered from the rejection alone.
+func validateContainerArgs(tool string, args map[string]any) error {
+	numberArg, forwarded := containerArgTools[tool]
+	if !forwarded {
+		return nil
+	}
+	cname, ok := requireString(args, "name")
+	if !ok {
+		return fmt.Errorf("missing required parameter: name")
+	}
+	if !docker.ValidName(cname) {
+		return fmt.Errorf("invalid container name: %s", cname)
+	}
+	if numberArg == "" {
+		return nil
+	}
+	lines := "50"
+	if v := stringArg(args, numberArg); v != "" {
+		lines = v
+	}
+	if !docker.ValidLines(lines) {
+		return fmt.Errorf("invalid line count: %s (must be a positive integer)", lines)
+	}
+	return nil
+}
+
 func (s *Server) executeTool(name string, args map[string]any) (any, error) {
 	if s.demo {
 		return s.executeDemoTool(name, args)
+	}
+
+	// Before any routing: whichever machine ends up answering, the argument
+	// rules are the same and were already checked.
+	if err := validateContainerArgs(name, args); err != nil {
+		return nil, err
 	}
 
 	if cap, ok := capabilityFor(name); ok && cap.supports(targetProxmox) {
--- a/internal/docker/docker.go
+++ b/internal/docker/docker.go
@@ -44,7 +44,7 @@ type ActionResult struct {
 }
 
 func Restart(name string) (*ActionResult, error) {
-	if !isValidName(name) {
+	if !ValidName(name) {
 		return nil, fmt.Errorf("invalid container name: %s", name)
 	}
 	out, err := util.DockerCmd("restart", name)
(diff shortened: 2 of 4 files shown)


internal/mcp/server.go
```diff
@@ -206,11 +206,63 @@ func (s *Server) handleToolCall(req *jsonRPCRequest) {
...
+}
+
 func (s *Server) executeTool(name string, args map[string]any) (any, error) {
 	if s.demo {
 		return s.executeDemoTool(name, args)
 	}
 
+	// Before any routing: whichever machine ends up answering, the argument
+	// rules are the same and were already checked.
+	if err := validateContainerArgs(name, args); err != nil {
```
Higangssh, line 259: The gate sits below the demo return, so demo mode is a third path that does not pass it. `docker_restart` with `name: "--help"` answers `invalid container name` on both real paths and canned success in demo.

Nothing is reachable through it — demo touches no docker binary and no remote host — so this is not a hole. I am raising it because the thesis of the PR is that one tool cannot answer differently depending on which way the call went, and demo is a way the call can go. A client exercising the surface against demo mode learns a validation contract that the real paths do not share.

Moving the gate above the demo branch would be one line. The argument for leaving it is that demo exists to show the shape of a successful response and rejecting input there teaches nothing. Either is defensible; I would rather it be chosen than inherited from statement order.

Higangssh (commented):
This is the shape I was hoping for, and it answers the design question in #83 rather than working around it.

Validating once in `executeTool` before the local-or-remote decision, while leaving the rules themselves in `internal/docker`, is the split that matters: the gate applies a rule it does not own, so there is still one definition of what a container name is. `#47` and `#65` were both about the opposite of that, and it would have been easy to fix this by writing a second copy of the check into `executeRemote`.

Three things I want to name specifically.

Keeping the checks inside `Restart`, `Stop`, `Logs`, `Top` and `Inspect` is right, and the reason you gave is the correct one — the CLI and alert remediation reach those functions without passing the MCP layer at all, so removing them would have traded one gap for another.

`ValidLines("")` returning false where the old inline loop passed it vacuously is the kind of change that usually breaks something quietly. It does not here, because every caller defaults to `"50"` first, and you said so rather than leaving me to work it out. Picking up `lines` at the same time closes the same shape I flagged in the issue — `--json` reaching the remote command as a flag rather than a bad number.

`TestForwardedContainerNamesHaveAGate` is the part that makes this hold. It reads `executeRemote`'s arms and cross-checks both directions, so a new tool forwarding `name` fails, and a stale map entry fails too. Source-parsing tests usually rot into false passes; this one `t.Fatal`s when its own parsing stops matching, with a message saying so, which is the property that makes the technique safe to use. `TestRegistryTargetsMatchRemoteArgvMapping` set the precedent and this follows it properly.

One inline comment, about demo mode being a third path. It is a question, not a blocker.

CI is green; `gofmt`, `go vet` and the `mcp`, `docker` and `cmd` package tests pass locally too.

lenny-ts:
hi, done

Higangssh:
Merged as 62d8d2c, and #83 closes with it. Thanks — six merged now.

Moving the gate above the demo branch was the right read of the question. The PR's whole argument is that a tool cannot answer differently depending on which way the call went, and demo was a way it could go.

One thing I did not ask for and would take as a follow-up if you feel like it: nothing fails if that ordering is undone. `TestForwardedContainerNamesHaveAGate` guards which tools are gated but not where the gate sits, so a later change that returns early for demo mode would pass. A test asserting a flag-shaped name is refused in demo mode would hold the four lines you just moved. Small, and entirely optional.

Issue #79, opened 2026-08-24
Proxmox endpoints are invisible in the web dashboard


Higangssh:
## Description

#78 adds read-only Proxmox visibility over the CLI and MCP. It does not touch `web/`, so a configured Proxmox endpoint appears nowhere in `homebutler serve`.

Phase 1 in #32 named the dashboard:

> Support JSON output and a concise cluster summary in the dashboard and MCP tools.

Deferring it is the right call and is agreed in #78 — the data shape should settle against real clusters before a second consumer is built on it. This issue is the named second phase that the deferral requires, so the gap does not go quiet once #78 merges.

## Scope

A read-only view, covering what the CLI already answers:

- Cluster name, quorum, and node online count — the `proxmox status` header
- Nodes with CPU, memory, and uptime
- Guests, QEMU and LXC together, with VMID, node, status, and name
- Storage usage

## What has to settle first

The typed shapes in `internal/proxmox/types.go`, once they have run against more than one cluster. Optional metrics, CPU fractions, and the boolean wire variants are the parts most likely to move.

How partial failure renders. A Proxmox endpoint is remote, so a token with limited ACLs returns some sections and not others. `inventory` settled this for local collection in #74 with `Warnings` and `Failed`; the dashboard needs the same distinction between "empty" and "not readable" rather than inventing a second convention for it.

## Notes

The dashboard is Svelte, so this is a different review axis from the Go client. Keeping it out of #78 keeps both reviewable.

Endpoint selection stays the CLI rule: optional when exactly one endpoint is configured, required when there is more than one. Proxmox endpoints do not join the SSH `--server` or `--all` fan-out.

## Done when

- `homebutler serve` shows a configured Proxmox endpoint without the user having to read the CLI
- A section that could not be read is distinguishable from one that is genuinely empty
- More than one endpoint is selectable

PR #90, opened 2026-08-26
feat: start, shut down, and reboot Proxmox guests with task status lookup

gsaraiva2109:
## Summary

Implements the Phase 2 write actions from #32: start/shutdown/reboot for QEMU and LXC guests, and single-task status lookup, exposed through the CLI and over MCP.

- Shutdown maps to Proxmox's graceful `status/shutdown`, not the hard `status/stop` power-off, per the maintainer's answer on #32: hard stop would make `stop` mean two different things across `homebutler docker stop` (graceful) and `homebutler proxmox guest stop`, and the destructive one would be the verb users reach for. The action→endpoint mapping is centralized in one small table so that decision is cheap to revisit.
- `--endpoint`/`--node`/`--type`/`--vmid`/`--confirm` (and the MCP equivalents) are all required and validated before token resolution or any network call — no unconfirmed action reaches the wire.
- The HTTP client refuses to follow redirects on POST, so a 307/308 can't cause a power action to be silently repeated against a different target.
- No automatic retry after an ambiguous POST failure (e.g. connection dropped mid-request).
- A successful action reports `accepted` with the opaque UPID, never completion — task status is a separate, explicit lookup that distinguishes running / stopped+OK / stopped+failed.

Reused Phase 1's `proxmox:` config, endpoint/token resolution, TLS handling, `targetProxmox`, and sanitized-fixture test approach. No new dependencies.

## Test plan

- [x] `go test ./internal/proxmox ./cmd ./internal/mcp` — 169 passed
- [x] `go test -race ./internal/proxmox ./cmd ./internal/mcp` — 169 passed
- [x] `go vet ./...`, `golangci-lint run`, `gofmt -l .` — clean
- [x] `go build ./...` — clean
- [x] `go test ./...` (full suite) — 1081 passed
- [x] Table-driven tests assert method, auth header, exact escaped path, and that invalid input (bad node/type/vmid/action, missing confirm) makes zero requests
- [x] Redirect-refusal and ambiguous-failure-no-retry covered by dedicated tests
- [x] MCP demo mode and registry-wide demo coverage updated for both new tools

### DIFF
--- a/internal/proxmox/client.go
+++ b/internal/proxmox/client.go
@@ -23,7 +23,12 @@ import (
 const defaultTimeout = 10 * time.Second
 
 const (
-	minVMID = 100
+	// minVMID matches the Proxmox API's own vmid schema minimum. 100 is only
+	// the default lower bound of /cluster/nextid's auto-assign range
+	// (datacenter.cfg next-id), not a floor the API enforces on an
+	// explicitly supplied vmid, and a cluster or an older pvesh-created
+	// guest can legitimately sit below it.
+	minVMID = 1
 	maxVMID = 999999999
 )
 
@@ -286,16 +291,16 @@ func ValidateTaskStatusRequest(node, upid string) error {
 	return nil
 }
 
-// TaskResult classifies the asynchronous task state for user-facing output.
+// TaskResult classifies the asynchronous task state as a short, stable token.
 func TaskResult(status, exitStatus string) string {
 	if status == "running" {
-		return "in flight"
+		return "running"
 	}
 	if status == "stopped" && exitStatus == "OK" {
-		return "completed successfully"
+		return "ok"
 	}
 	if status == "stopped" {
-		return "completed with failure"
+		return "failed"
 	}
 	return "unknown"
 }
--- a/internal/proxmox/client.go
+++ b/internal/proxmox/client.go
@@ -1,4 +1,4 @@
-// Package proxmox provides the small read-only Proxmox VE API client used by
+// Package proxmox provides the small Proxmox VE API client used by
 // HomeButler's Proxmox commands.
 package proxmox
 
@@ -22,6 +22,17 @@ import (
 
 const defaultTimeout = 10 * time.Second
 
+const (
+	minVMID = 100
+	maxVMID = 999999999
+)
+
+var guestActionPaths = map[GuestAction]string{
+	GuestActionStart:    "start",
+	GuestActionShutdown: "shutdown",
+	GuestActionReboot:   "reboot",
+}
+
 // Options configures a Proxmox API client. Token is the resolved token value;
 // resolving token files belongs to the configuration layer.
 type Options struct {
@@ -35,7 +46,7 @@ type Options struct {
 	Timeout     time.Duration
 }
 
-// Client is a read-only client for the Proxmox VE JSON API.
+// Client is a client for the Proxmox VE JSON API.
 type Client struct {
 	baseURL string
 	tokenID string
@@ -76,6 +87,12 @@ func New(options Options) (*Client, error) {
 		http: &http.Client{
 			Timeout:   options.Timeout,
 			Transport: &http.Transport{TLSClientConfig: tlsConfig},
+			CheckRedirect: func(_ *http.Request, via []*http.Request) error {
+				if len(via) > 0 && via[0].Method == http.MethodPost {
+					return http.ErrUseLastResponse
+				}
+				return nil
+			},
 		},
 	}, nil
 }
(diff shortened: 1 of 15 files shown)


internal/proxmox/types.go
```diff
@@ -250,6 +250,30 @@ type Task struct {
...
+	Node       string `json:"node"`
+	PID        int64  `json:"pid,omitempty"`
+	PStart     int64  `json:"pstart,omitempty"`
+	StartTime  int64  `json:"starttime,omitempty"`
+	Type       string `json:"type"`
+	ID         string `json:"id,omitempty"`
+	User       string `json:"user,omitempty"`
+	Status     string `json:"status"`
+	ExitStatus string `json:"exitstatus,omitempty"`
+	Result     string `json:"result"`
```
Higangssh, line 274: `result` carries English prose — `in flight`, `completed successfully`, `completed with failure` — into a field that 1.0 freezes. A consumer that wants to branch on the outcome has to match a sentence, and a later reword of that sentence is then a breaking change to the schema rather than a copy edit.

The repository's own convention is a short token: `docker.ActionResult` reports `Status: "ok"`. Something like `running` / `ok` / `failed` / `unknown` keeps the derived field useful to a caller and leaves the phrasing to whoever renders it for a person.

The distinction is the one #74 settled for `inventory`: `Warnings` carries the human-readable half and `Failed` the machine-readable one, kept separate so neither has to do both jobs. Here they are the same field.

internal/proxmox/client.go
```diff
@@ -22,6 +22,17 @@ import (
 
 const defaultTimeout = 10 * time.Second
 
+const (
+	minVMID = 100
```
Higangssh: Worth confirming this floor is Proxmox's rule rather than the web UI's convention. The GUI will not create a guest below 100, but a VMID assigned through `pvesh` or an older cluster may sit under it, and `ValidateGuestAction` refuses those before a request is made.

The part that makes it awkward is the asymmetry with Phase 1: `proxmox guests` has no VMID floor, so it will list a guest at VMID 50 that `proxmox guest shutdown` then declines to touch. A tool that shows you something it will not act on is a worse failure than one that tries and gets a Proxmox error back naming the real constraint.

If 100 is genuinely the API minimum this is right as written and only wants a comment saying so. If it is a UI convention, letting Proxmox reject the VMID gives the operator the accurate reason.

Higangssh (commented):
Both decisions from #32 landed, and the tool split is exactly the shape I asked for:

| Tool | Risk |
| --- | --- |
| `proxmox_guest_start` | `riskWrite` |
| `proxmox_guest_reboot` | `riskWrite` |
| `proxmox_guest_shutdown` | `riskDestructive` |
| `proxmox_task_status` | `riskRead` |

That is the same line `docker_restart` and `docker_stop` already draw, so a client that gates destructive tools can now allow powering a guest on without also allowing it to be shut down.

Three things beyond the ask that I want to name, because they are the parts that would have been easy to skip.

`--confirm` really is checked before credentials are resolved — `ValidateGuestAction` and the confirmation gate both run ahead of `openProxmoxClient`, so an unconfirmed action never loads a token, let alone reaches the network. The error repeats the full resolved target and the exact command to rerun, which is the shape the rest of the CLI uses.

Refusing to follow a redirect whose original request was a POST is a good catch, and not one I would have asked for. A 307 on a power action is a request to perform it somewhere else, and a client that quietly obliges has repeated a side effect against a target the operator never named. Pairing it with no automatic retry after an ambiguous failure covers the other half of the same problem.

Reporting `accepted` with the UPID rather than claiming the action finished is the right call, and the separate task lookup that distinguishes running from stopped-OK from stopped-failed is what makes it usable rather than merely honest.

Two inline comments. The `result` field is the one I would like changed before this merges, because 1.0 freezes the JSON schema and that field currently holds a sentence. The VMID floor is a question rather than a finding.

**CI has not run on this branch** — no workflow run and no check suite exist for `dd3f794`, which is different from #88 on the same day. That needs sorting before merge, since `main` requires those checks. Locally `gofmt`, `go vet`, `go build ./...` and `go test ./...` are all clean on this branch.

One small thing in the PR body: it says #32 "doesn't record an explicit maintainer choice" between `shutdown` and `stop`. It does now — I answered about ninety minutes before you opened this. The code already matches that answer, so only the sentence is stale, but it is the sentence a future reader will use to work out why the graceful operation was chosen.

Higangssh:
CI is green now — lint, tests and all four build targets. No workflow run had been created for the branch at all, which is why the checks were blank; reopening the PR fired the `reopened` trigger and it ran normally. Nothing on your side, and nothing you need to do about it.

That leaves the `result` field as the one thing I would like changed before this merges. The VMID floor is still a question rather than a request.

gsaraiva2109:
Pushed 3b49895.

**`result`** is now a short token — `running` / `ok` / `failed` / `unknown` — matching `docker.ActionResult`'s `Status` convention. Updated the two call sites, both packages' tests, and the CLI docs paragraph that spelled out the old prose.

**VMID floor**: checked, and you had it right — 100 is not an API minimum. It's the default lower bound of `/cluster/nextid`'s auto-assign range (`datacenter.cfg next-id`), which only governs auto-assignment; the schema for an explicitly-supplied `vmid` parameter doesn't enforce it, and a cluster or an older `pvesh`-created guest can sit below it. That's also exactly the asymmetry you pointed out — `proxmox guests` would list one, `proxmox guest shutdown` would refuse it. Lowered `minVMID` to 1 (matching the actual API schema minimum) and let Proxmox return its own error for anything genuinely invalid, updated the MCP tool description and the boundary test to match.

Also fixed the stale sentence in the PR description — it now points at your #32 answer instead of saying the choice wasn't recorded.

`go vet`, `golangci-lint run`, `gofmt -l .`, `go build ./...`, and `go test ./...` (1081, full suite) are all clean on the branch.

Higangssh:
Merged as 309c35a.

The VMID question is the one I want to record, because you turned it into something better than what I asked. I raised it as a hunch — the floor looked like a UI convention and the asymmetry with `proxmox guests` bothered me — and I had no idea whether 100 was real. You went and found out that it is `datacenter.cfg next-id`, governing auto-assignment and not the schema for an explicitly supplied vmid, then fixed the code, the tool description and the test, and wrote the reason into the comment so the next person does not have to repeat the lookup. A question answered with a citation beats a question answered with a change.

The redirect refusal on POST is the part I would not have thought to ask for. A 307 on a power action is an instruction to perform it somewhere else, and following it silently repeats a side effect against a target nobody named.

Phase 2 is done. #32 stays open for Phase 3; the design question there is #62, which is the one worth settling before any of it gets written, and your operating experience is most of what that thread is still missing. The dashboard is #79 whenever the data shape has met a second cluster.

PR #94, opened 2026-08-27
feat: print a Proxmox Community Script command instead of running one

gsaraiva2109:
## Summary

Closes the narrowest slice of #62: whether homebutler runs externally-authored code at all. It doesn't — this generates and prints the install command for a curated Proxmox VE Community Script and stops there. The operator reviews it and runs it by hand on the Proxmox host.

- `internal/proxmox/scripts.go`: a small curated catalog (`docker`, `homeassistant`, `nginxproxymanager`, `pihole`, `postgresql`) with `Scripts()` and `ScriptCommand(slug)`. The rendered command is pinned to one commit of `community-scripts/ProxmoxVE` (`fadcb0c3...`) rather than tracking `main`, so the bytes a script fetches today match a week from now — addresses the pinning question #62 raised. Bumping the pin is a deliberate, reviewable one-line change.
- CLI: `proxmox script list`, `proxmox script show <slug>`.
- MCP: `proxmox_script_list`, `proxmox_script_command` (both `riskRead`, `targetLocal` — no endpoint or credentials involved, since nothing is executed or contacted).

## Notes for reviewers

- No execution path was added anywhere — `ScriptCommand` only formats a string. `proxmox_guest_start/reboot/shutdown` (Phase 2, already merged) remain the only guest-affecting Proxmox actions.
- Catalog and pinned ref checked live against `community-scripts/ProxmoxVE` (`gh api`) rather than invented.
- #62's remaining questions — whether homebutler ever executes such a script, and what that needs from the permission model — are untouched by this PR.

## Test plan

- [x] `gofmt -w .`, `go vet ./...`, `golangci-lint run`, `go test ./...` (1092 tests), `go build ./...`
- [x] Manually ran `proxmox script list` / `proxmox script show docker|nope` (human + `--json`)
- [x] Exercised `proxmox_script_command` through the real MCP server (`mcp --demo`) over stdio

### DIFF
--- a/cmd/proxmox.go
+++ b/cmd/proxmox.go
@@ -314,9 +314,9 @@ func newProxmoxScriptShowCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			result := proxmoxScriptCommandResult{Slug: args[0], Command: command}
+			result := proxmoxScriptCommandResult{Slug: args[0], Command: command, Warning: proxmox.ScriptWarning}
 			return writeProxmox(cmd, result, jsonOutput, "Community Script command", func(b *strings.Builder) {
-				fmt.Fprintf(b, "Slug: %s\nCommand: %s\n\nReview it, then run it yourself on the Proxmox host; homebutler does not run it for you.\n", result.Slug, result.Command)
+				fmt.Fprintf(b, "Slug: %s\nCommand: %s\n\n⚠️  %s\n\nReview it, then run it yourself on the Proxmox host; homebutler does not run it for you.\n", result.Slug, result.Command, result.Warning)
 			})
 		},
 	}
@@ -326,6 +326,7 @@ func newProxmoxScriptShowCmd() *cobra.Command {
 type proxmoxScriptCommandResult struct {
 	Slug    string `json:"slug"`
 	Command string `json:"command"`
+	Warning string `json:"warning"`
 }
 
 type proxmoxGuestActionResult struct {
--- a/cmd/proxmox.go
+++ b/cmd/proxmox.go
@@ -17,7 +17,7 @@ func newProxmoxCmd() *cobra.Command {
 		Short:        "Inspect Proxmox VE endpoints",
 		SilenceUsage: true,
 	}
-	cmd.AddCommand(newProxmoxStatusCmd(), newProxmoxGuestsCmd(), newProxmoxNodeCmd(), newProxmoxTasksCmd(), newProxmoxGuestCmd(), newProxmoxTaskCmd())
+	cmd.AddCommand(newProxmoxStatusCmd(), newProxmoxGuestsCmd(), newProxmoxNodeCmd(), newProxmoxTasksCmd(), newProxmoxGuestCmd(), newProxmoxTaskCmd(), newProxmoxScriptCmd())
 	return cmd
 }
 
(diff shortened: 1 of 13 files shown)


cmd/proxmox.go
```diff
@@ -282,6 +282,52 @@ func newProxmoxTaskCmd() *cobra.Command {
...
+		Short: "Print the install command for one Community Script (does not run it)",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			command, err := proxmox.ScriptCommand(args[0])
+			if err != nil {
+				return err
+			}
+			result := proxmoxScriptCommandResult{Slug: args[0], Command: command}
+			return writeProxmox(cmd, result, jsonOutput, "Community Script command", func(b *strings.Builder) {
+				fmt.Fprintf(b, "Slug: %s\nCommand: %s\n\nReview it, then run it yourself on the Proxmox host; homebutler does not run it for you.\n", result.Slug, result.Command)
```
Higangssh: This is the line I care most about in the PR, and it is not saying enough yet.

What #62 asked for was that the output state two things: that the script is **not reviewed by homebutler**, and that it **runs as root** on the Proxmox host. What is here says homebutler does not run it, which a reader takes as a note about who types the command rather than a warning about what the command does.

The reason it matters is the one that made the decision go this way at all: a command homebutler formats carries homebutler's authority. "The tool told me to run this" is the risk, and the output has to work against that rather than lean on it. Something closer to:

```
⚠️  This script is not reviewed by homebutler. It is fetched from
    community-scripts/ProxmoxVE and runs as root on the Proxmox host.
```

Two words are doing the work there — *not reviewed* and *as root* — and neither is present today.

cmd/proxmox.go
```diff
@@ -282,6 +282,52 @@ func newProxmoxTaskCmd() *cobra.Command {
...
+			})
+		},
+	}
+	return cmd
+}
+
+type proxmoxScriptCommandResult struct {
```
Higangssh, line 326: The warning does not exist at all outside the human renderer:

```
$ homebutler proxmox script show docker --json
{"slug":"docker","command":"bash -c \"$(curl -fsSL https://.../ct/docker.sh)\""}
```

`proxmox_script_command` returns the same shape over MCP. So an agent receives a root-level shell command with nothing marking it as unreviewed third-party code, and relays it to a person with homebutler's name attached — which is the exact failure the print-don't-execute decision was meant to avoid, arriving through the channel that decision was most concerned about.

A `warning` field on this struct rides along in both `--json` and the MCP result and costs one line. #74 made the same split for a different reason: the human-readable half and the machine-readable half both have to carry the information, because a caller only sees one of them.

Higangssh (commented):
The decision in #62 is implemented faithfully, and the part I would have checked first is the part you got right: there is no execution path anywhere. `ScriptCommand` formats a string and returns it, and nothing in the package fetches or runs anything. That is the whole point of the decision and it would have been easy to smuggle a convenience `--run` flag in alongside it.

Pinning to `fadcb0c3` rather than `main` is what #62 asked for and the comment explains why the pin exists, which is what makes a future bump a deliberate act rather than a routine one. Checking the catalog and the ref against the real repository instead of writing plausible-looking slugs is the kind of thing that only shows up later, when someone pastes a command that 404s.

`riskRead` and `targetLocal` are both right. It contacts nothing and it cannot be pointed at an endpoint, so it is not a `targetProxmox` tool despite the name.

**Two inline comments, and the first is a request.** The warning text does not say the two things #62 asked it to say, and it does not exist at all in `--json` or over MCP. That warning is not decoration around the feature; on the MCP path it is most of the feature. I would like both addressed before this merges.

**Docs are untouched.** #78 and #90 both updated `docs/proxmox.md`, and this adds two commands and two MCP tools without a line anywhere. `proxmox script list` and `proxmox script show` need a section, and it is the natural place to write down why homebutler stops at printing — a user who finds this command and wonders why it does not just run the thing should not have to read #62.

While you are in there: `docs/mcp-server.md`'s tool table has `docker_top` and `watch_history` but no Proxmox tools at all, and `skills/SKILL.md` lists none either. That drift started in #78 and continued through #90, so it is not yours — but eleven tools missing from the table is worth one pass now rather than a bigger one later. Say the word if you would rather I do that separately.

One thing nobody needs to act on yet: nothing checks that the pinned ref actually contains the catalogued scripts. If a bump lands after an upstream rename, `ScriptCommand` prints a URL that 404s, `curl -f` fails, and `bash -c "$()"` runs an empty string — so it fails quietly rather than dangerously, but the operator sees nothing happen and no reason why. Not worth a network test; worth a sentence in whatever bump procedure ends up in the docs.

CI is green, and `gofmt`, `go vet` and `go test ./...` pass locally.

Higangssh:
Merged as c58b096.

Pulling the warning into `ScriptWarning` was better than what I asked for. I described a field on one struct; a single constant means the human renderer, `--json`, the MCP result and demo mode cannot drift apart, and the next path that renders a command gets the warning by using the same name rather than by remembering to. The comment on it says which two facts it exists to carry, so a future edit that softens the wording has to argue with the reason.

You also fixed the MCP tool table, which I said was not yours. Eleven tools had been missing from it since #78 and it would have kept growing.

That closes the question #62 was opened for. Phase 3 exists and homebutler does not execute anything. #32 stays open — what remains there is whether execution ever becomes worth revisiting, and that is a separate conversation with a precondition attached rather than an item of work.

PR #95, opened 2026-08-27
feat: show Proxmox status in web dashboard

### LINKED ISSUE #79: Proxmox endpoints are invisible in the web dashboard

gsaraiva2109:
## Problem

`homebutler serve` leaves configured Proxmox endpoints invisible even though the CLI and MCP expose the read-only cluster view. Dashboard users must switch interfaces, and a restricted token can look like an empty environment when collectors cannot read part of the API.

## Change

The dashboard now adds a Proxmox VE card that:

- Lists configured endpoints separately from SSH servers.
- Selects the sole endpoint automatically and shows a selector when multiple endpoints exist.
- Displays the cluster name, quorum, online node count, node CPU, memory and uptime, unified QEMU and LXC guests, and storage usage.
- Renders collector failures and warnings as unavailable states instead of empty data.

The server adds authenticated, CORS-wrapped endpoints for safe endpoint names and selected status. It reuses `Config.SelectProxmox`, the existing TLS and token configuration, and `proxmox.DefaultView`. Endpoint discovery returns names only. Credential and certificate file errors are logged server-side and reduced to generic API errors.

An entirely empty `/cluster/resources` response is marked as a resource collector failure because a connected cluster always has at least one node. This identifies ACL-filtered responses while allowing guest and storage sections to be genuinely empty when nodes remain visible.

**Behaviour change beyond the dashboard:** `DefaultView` backs `proxmox status`, the `proxmox_status` MCP tool, and the new dashboard endpoint alike, so this also changes `proxmox status --json` and `proxmox_status`. A cluster whose token is ACL-limited to the point where `/cluster/resources` returns no nodes, guests, or storage at all now reports `failed_collectors: ["resources"]` instead of three empty lists. CLI and MCP consumers parsing `failed_collectors` should account for this.

## Scope

The dashboard remains read-only and does not expose guest actions. Proxmox endpoints remain separate from SSH `--server` and `--all` fan-out. This adds no caching, polling, or new Proxmox API calls.

### Prerequisites

- A Proxmox endpoint already configured under `proxmox:`.
- A dedicated API token with the read-only `PVEAuditor` role.
- The token value stored in a `0600` file and referenced with `token_file`.
- TLS configured with a pinned `fingerprint` or `ca_file`. Use `insecure: true` only as a temporary test fallback.

See [proxmox.md](docs/proxmox.md) for the complete least-privilege and TLS setup.

### Build and start

```bash
make build-all
./homebutler config validate
./homebutler serve --host 127.0.0.1 --port 8080
```

Open <http://127.0.0.1:8080>. Keep the terminal visible so connection or token errors are easy to correlate with the page.

### Healthy endpoint

With one endpoint configured, verify that:

- The **Proxmox VE** card appears without an endpoint dropdown.
- The badge shows the configured endpoint name.
- The cluster name, quorum state, and online/total node count match Proxmox.
- Every node shows status, CPU percentage, memory usage, and uptime.
- QEMU and LXC guests share one table and show VMID, name, type, node, and status.
- Storage shows its name, node, status, and used/total bytes.
- Missing optional metrics show `—`; a measured zero shows `0.0%` or `0 B`.

Compare the card with the existing CLI response:

```bash
./homebutler proxmox status --json
```

The browser and CLI should describe the same cluster snapshot. Small metric differences are expected if the two requests run at different times.

### Multiple endpoints

Configure a second Proxmox entry with a different `name`, then restart `homebutler serve`.

- The card should show a labeled endpoint dropdown.
- Selecting either endpoint should replace all cluster, node, guest, and storage data.
- Switching quickly must not let a slower response overwrite the currently selected endpoint.
- The SSH server dropdown must not contain Proxmox endpoints.

Compare each selection explicitly:

```bash
./homebutler proxmox status --endpoint pve-a --json
./homebutler proxmox status --endpoint pve-b --json
```

### Empty versus unreadable

These are intentionally different states.

1. On a readable endpoint with no guests or no visible storage, verify the relevant section says `No guests found` or `No storage found`.
2. Create a separate test token with deliberately limited ACLs. Do not weaken the production token. Select that endpoint and verify:
   - A failed cluster collector says `Cluster status unavailable`.
   - A failed resources collector says node, guest, and storage inventory is unavailable, never empty.
   - Sections that remain readable still render normally.
   - The warning list explains which collector failed.
3. Use an unreachable test endpoint and verify the card reports unavailable collectors and warnings without showing an empty healthy cluster.

The machine-readable response can be checked directly:

```bash
curl -sS 'http://127.0.0.1:8080/api/proxmox/status?endpoint=pve-a'
```

A partial result should remain HTTP 200 and include both `warnings` and `failed_collectors`. A successfully read empty collection must not list that collector in `failed_collectors`.

### Security and accessibility

- Open `/api/proxmox/endpoints` and verify it contains endpoint names only. It must not expose token values, token IDs, token-file paths, CA paths, or fingerprints.
- Navigate to the endpoint selector with the keyboard, change it with arrow keys, and confirm the data updates.
- Verify loading, error, empty, and unavailable states use text rather than color alone.
- Test at 200% browser zoom and a narrow mobile viewport. Tables may scroll horizontally, but the page must not lose data or controls.
- If `serve --token` is part of your normal setup, repeat the API checks with the expected `Authorization: Bearer ...` header and confirm unauthenticated requests return HTTP 401.

Closes #79


internal/proxmox/client.go
```diff
@@ -323,6 +323,9 @@ func (c *Client) DefaultView(ctx context.Context) (DefaultView, error) {
 	if resources, err := c.Resources(ctx); err != nil {
 		view.Warnings = append(view.Warnings, "resources: "+err.Error())
 		view.Failed = append(view.Failed, CollectorResources)
+	} else if len(resources.Nodes) == 0 && len(resources.Guests) == 0 && len(resources.Storage) == 0 {
+		view.Warnings = append(view.Warnings, "resources: no resources visible; check Proxmox token permissions")
```
Higangssh, line 327: This changes `proxmox status` and the `proxmox_status` MCP tool, not only the dashboard. `DefaultView` is shared by all three, so a CLI user whose token is ACL-limited now gets `failed_collectors: ["resources"]` where they previously got empty lists.

I think the change is right. A cluster that answers `/cluster/resources` at all has at least one node, so all three lists empty is a permission result rather than a real one, and reporting it as unavailable is more honest than printing `0 nodes | 0 guests | 0 storage`.

Two things follow from it being right rather than dashboard-local:

`Failed` has meant "this collector did not answer" since #74. It now also means "answered, but the answer cannot be true". That is a reasonable widening and the field's doc comment should say so, or the next person reading it will assume a transport error.

The PR describes this under the dashboard's rendering and the Scope section lists what the PR does not do without mentioning it. It is a behaviour change to `--json` output that a release note has to carry, so it is worth stating plainly in the description that CLI and MCP consumers see it too.

docs/proxmox-dashboard-testing.md
```diff
@@ -0,0 +1,113 @@
+# Testing the Proxmox dashboard
+
+Use this guide to test the implementation of GitHub issue #79 before opening a
```
Higangssh: This reads as contributor workflow rather than user documentation — it opens with "before opening a pull request", references issue #79, and walks through `git fetch fork`. Everything else in `docs/` is written for someone using homebutler: `backup.md`, `configuration.md`, `mcp-server.md`, `proxmox.md`.

The content is worth having. A dashboard is the part of this repository unit tests reach least, and a written manual pass is the substitute. But shipping it under `docs/` puts instructions for reviewing an unmerged branch in front of users.

The PR description is where a reviewer needs it, and CONTRIBUTING is where it belongs if it should outlive the PR. Either way I would rather it not ship in the released documentation set.

Higangssh (commented):
Closes the last piece of #32's original Phase 1 scope, and does it after the data shape had a chance to settle, which is what #79 was created to wait for.

I checked the thing I was most likely to find wrong and did not: the two new routes go through the same `api()` wrapper as everything else, so `requireAuth` and `cors` both apply. Registering them outside the demo/non-demo branch rather than duplicating them reads differently from the routes around it, but the handlers take demo into account themselves and the result is less to keep in sync, so I would leave it.

Demo mode does the right thing without a special case in the frontend: `handleProxmoxEndpoints` returns an empty list, and the card's `{#if}` hides it rather than rendering an error next to the screenshots.

Returning only endpoint names, and reducing credential and certificate failures to a generic message with the detail logged server-side, is the correct shape for a dashboard endpoint. It would have been easy to pass the configuration error straight through.

Two inline comments. Neither blocks, but the first should change the PR description before this merges, because it is a `--json` behaviour change that CLI and MCP consumers inherit and a release note has to catch it.

CI is green, and `gofmt`, `go vet` and `go test ./...` pass locally.

One thing for after this lands rather than in it: the README still has no mention of Proxmox at all — not the commands from #78, #90 or #94, and now not the dashboard either. `docs/proxmox.md` is good and there is no path to it from the front page. I am picking that up separately; nothing for you to do.

Higangssh:
Merged as 504ac1e, and #79 closes with it. That is Proxmox complete — #78, #90, #94 and now the dashboard.

Both of my comments came back better than I asked for them. I said the `Failed` doc comment should note that the meaning had widened; you wrote down what the widened case actually is and what it points at, so the next person reading that field does not have to guess whether an entry means a transport error. And I asked for the behaviour change to be stated plainly in the description — you named the two tools it reaches, the exact field, and what a consumer parsing it should do. The release note for this can be lifted straight out of that paragraph, which is not usually true of a PR description.

Worth recording what this run looked like from here, since four PRs is a lot to have landed in a week. You checked the Proxmox API rather than assuming it on the VMID floor and came back with the actual reason. You refused a redirect on POST without being asked, which is the sort of thing that only gets noticed after it has already caused an incident somewhere. You asked about the dashboard rather than quietly dropping it, and about `stop` versus `shutdown` rather than picking one and letting a reviewer find it. Every one of those was you deciding that a question was cheaper than a correction.

The one thing I would still like, whenever it suits you: #62 is closed on the reasoning that homebutler does not execute externally-authored code, and it records what would have to be true for that to be revisited — an allowlist and a pin the operator writes themselves. If Phase 3 ends up being something you use often enough to be annoyed by the copy-paste, that is the signal, and I would rather hear it from someone running it than guess.

Thanks. This was a good week of review.

Issue #91, opened 2026-08-26
restore: bind-mount symlink can escape --allow-bind containment


gsaraiva2109:
## Summary

`bindTargetAllowed` in `internal/backup/restore.go` (around line 53) only does a lexical string-prefix check against the `--allow-bind` root. A hostile archive can use one bind mount to plant a symlink under the allowed root, then use a second bind mount in the same restore to extract *through* that symlink, writing outside the allowed root entirely — defeating the containment added for GHSA-v8mc-vpp8-jr4p / CWE-22 / CWE-73 in 0.22.1.

## Reproduction sketch

Given `--allow-bind /srv/app`:

1. Archive declares bind mount **A** with `Source=/srv/app/shared`, whose payload contains one entry that is a symlink `foo -> /etc`.
2. Archive declares bind mount **B** with `Source=/srv/app/shared/foo`.
3. `bindTargetAllowed` passes both — both are lexically under `/srv/app`.
4. Restoring **A** creates the symlink at `/srv/app/shared/foo`.
5. Restoring **B**: `restoreMount`'s `os.MkdirAll(m.Source, ...)` no-ops on the existing symlink (`Stat` follows it, sees a directory; `Lstat` still shows it as a symlink). The subsequent `tar xzf B.tar.gz -C /srv/app/shared/foo` follows the symlink via `chdir` and extracts B's payload into `/etc`.

GNU tar's own unsafe-symlink protection only guards against escapes *within* a single tar invocation — it doesn't help when `-C` itself already points through a symlink planted by an earlier mount in the same restore.

## Suggested direction

`bindTargetAllowed` (or `restoreMount` just before extraction) should resolve the target with something equivalent to `Lstat` + reject if any path component under the allowed root is a symlink, rather than trusting a purely lexical prefix check. Extraction targets should be validated after symlink resolution, not before.

## Context

Found during review of #90 (Proxmox Phase 2), while checking test coverage in the area `restore.go` touches. Unrelated to that PR — `restore.go` isn't part of its diff — filing separately.

Higangssh:
Confirmed, and reproduced end to end on Linux — thank you for writing this up properly. The two-mount construction is the part that makes it serious: I had looked at the same lexical check in #89 and filed it as hardening, because I assumed the symlink had to already be sitting on the host. It doesn't. The archive plants it and walks through it within a single restore, and neither mount looks unusual on its own.

[...]

PR #112, opened 2026-08-30
fix: restore trusted tar to decline the archive members that escape the destination

### LINKED ISSUE #87: restore relies on tar's own refusal to keep a payload inside the directory it validated
## Description

GHSA-v8mc-vpp8-jr4p fixed which directory `restore` writes into. It did not change how the payload gets written once a directory is chosen:

```go
_, err := util.RunCmd("tar", "xzf", archivePath, "-C", m.Source)
```

`archivePath` is a tarball from inside the backup archive, so its member names are chosen by whoever built it, exactly as the manifest paths were. A member named `../../etc/something`, or an absolute path, is asking to be written outside the directory we validated.

## Why nothing is broken today

Both tar implementations homebutler runs against refuse this on their own. GNU tar strips a leading `/` and skips members containing `..`; bsdtar does the equivalent. So the validated destination holds.

The problem is that it holds because of a default in a program we shell out to, not because of anything in this repository. Nothing here asserts it, no test covers it, and a host with a tar that behaves differently — or a future switch to an archive library — would lose the property without any signal.

## Scope

`restoreMount` (`internal/backup/restore.go`), and the same `tar xzf -C` shape in `internal/backup/drill.go:355`. Drill extracts into an isolation directory, so an escape there reaches less, but it is the same construction.

Options, roughly in order of how much they change:

- Pass the flags that make the refusal explicit rather than default, where they are portable across GNU tar and bsdtar
- List the archive first (`tar tzf`) and refuse any member that is absolute or contains a `..` component, before extracting
- Extract through Go's `archive/tar`, which puts the check in code that can be tested

The middle option is probably the best trade: it is a few lines, it works the same on both tars, and it turns an inherited default into an assertion.

## Done when

- A payload member that tries to leave the destination is refused by homebutler, not by tar
- A fixture archive containing such a member is in the test suite
- `drill` gets the same treatment or a written reason why it does not need it

## Notes

Raised while fixing GHSA-v8mc-vpp8-jr4p and deliberately left out of that fix, which was already touching enough. Not known to be exploitable on a stock GNU or BSD tar.

### LINKED ISSUE #92: restore: archive-path derivation duplicated across three call sites
## Summary

In `internal/backup/restore.go`, `mountHasPayload`, the bind branch of `refuseMount`, and `restoreMount` each independently recompute the per-mount archive path as `filepath.Join(volDir, sanitizeName(m.Name)+".tar.gz")`. Three separate implementations of the same derivation.

Not a bug today, but a maintenance hazard: if `sanitizeName`'s contract ever changes (for example, to also gate on `Type`, the way bind-mount handling effectively keys off `m.Name`), one of the three call sites could silently drift from the others and reopen a gap in the containment these functions exist to enforce.

## Suggested direction

Extract a single `mountArchivePath(m, volDir) string` helper and have all three call it, so the three checks stay provably consistent by construction.

## Context

Found during review of #90 (Proxmox Phase 2), while checking test coverage in the area `restore.go` touches. Unrelated to that PR — filing separately. Low priority; bundle with #91 if convenient since both touch the same functions.

Higangssh:
Closes #87. Closes #92.

`restore` and `backup drill` shelled out to `tar xzf -C <dir>`. That was safe, but for a reason that lives outside this repository: GNU tar and bsdtar strip a leading `/` and skip members containing `..` on their own. Member names come from whoever built the archive, exactly as manifest paths do, so the destination was being held by a default in a program we invoke. Nothing here asserted it, no test covered it, and a host with a different tar — or a later move to an archive library — would have lost it with no signal.

Extraction now happens in `internal/backup/extract.go`, and refuses four things:

- an absolute member name
- a member that climbs out of the root with `..`
- a hard link naming a file outside the tree, which never spells a path outside it but hands the extracted tree a second name for one
- a member written through a symlink an earlier member of the same archive planted — #91 one level down, where every component is relative and none contains `..`, so a name check alone accepts it

**Refusing rather than skipping is the behaviour change worth arguing about.** Skipping is what tar does, and it means an archive that tried to escape still reports success having restored less than it claimed. A bind target that was never permitted stays a reported-and-skipped refusal, because that is an operator configuration answer they can act on. A member that lies about where it belongs is the archive being untrustworthy, and continuing to read from it is not the safe response. Mounts already restored before the failure stay restored.

The extractor keeps what tar was giving us, and one of those is easy to lose: `sudo homebutler restore` is documented in `docs/backup.md` for the case where the target is not writable, and tar restores ownership when it runs as root. A Go rewrite that skipped that would silently break the restores most likely to need it. Directory modes are applied last for the same reason tar makes a second pass — a directory stored `0500` must not block its own contents.

The volume branch still extracts inside an alpine container into a Docker volume and is deliberately untouched: the destination there is the volume, not a host path.

Verified end to end on Linux (aarch64), unprivileged and as root, rather than only in unit tests:

```
round trip, unprivileged  → mode=600  mtime=2020-03-04  symlink intact
                            locked dir mode=500 with its contents written
round trip, uid 0         → owner=1000:1003, not 0:0
planted symlink escape    → error: refusing member "hop/canary":
                              /tmp/evil87/dest/hop resolves to /tmp/victim87,
                              outside the destination
                            exit 1, canary = ORIGINAL
```

#92 rides along because it is the same functions: `mountHasPayload` decides whether the containment check runs at all and `restoreMount` decides what gets written, so those two agreeing about which archive a mount refers to is load-bearing. They agreed by identical code in three places, one of them in a different file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### 🔐 Security
+
+- check archive member names in homebutler rather than relying on `tar` to decline them (#87). `restore` and `backup drill` extracted with `tar xzf -C <dir>`, which held the destination only because GNU tar and bsdtar strip a leading `/` and skip members containing `..` of their own accord. Nothing here asserted it and no test covered it, so a host with a different tar would have lost the property with no signal. Extraction is now done in homebutler, and a member that is absolute, that climbs out of the root, that is a hard link to a file outside the tree, or that is written through a symlink an earlier member of the same archive planted, fails the restore instead of being skipped
+
+### ♻️ Refactoring
+
+- derive a mount's archive path in one place (#92). `mountHasPayload`, `restoreMount` and `backup drill` each rebuilt `volDir/sanitizeName(name).tar.gz` independently, and the first two agreeing is load-bearing: `mountHasPayload` decides whether the containment check runs at all, and `restoreMount` decides what gets written. They agreed by identical code rather than by construction
+
+### ⚠️ Behavior changes
+
+- **An archive member that would be written outside its destination now fails the restore**, where `tar` skipped it and reported success. A restore that was quietly dropping members will now stop and name the member it refused. Mounts already restored before the failure stay restored — a mount target that was never permitted is still reported under `refused` and skipped, because that is an operator configuration answer; a member that lies about where it belongs is the archive being untrustworthy, and continuing to read from it is not the safe response
+
+### 🧪 Tests
+
+- cover the four escapes the extractor now refuses, and the behaviour it has to preserve: file modes, modification times, symlinks, ownership when running as root, and a directory stored read-only whose contents must still be written into it
+
 ## [0.23.1](https://github.com/Higangssh/homebutler/compare/v0.23.0...v0.23.1) - 2026-08-30
 
 **`--allow-bind` could be walked out of by the archive it was meant to contain.** The check that a bind target sits inside a permitted root compared paths as text, and text does not follow symlinks. One archive with two bind mounts defeats it without needing anything to exist on the host beforehand: the first restores normally inside the permitted root and its payload plants a symlink there, the second names that symlink as its source, and the extraction follows it wherever it points. Upgrade if you have ever run `homebutler restore --allow-bind` on an archive from anywhere but your own machine.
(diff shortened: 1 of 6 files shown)


Higangssh:
Verified against a backup homebutler produced itself, rather than only against archives built for the tests.

A real compose project on Linux (aarch64) with a named volume and a bind mount, `homebutler backup`, then the originals destroyed — bind directory emptied, `docker compose down -v` to remove the volume — and `homebutler restore` run with this branch:

```
file list      identical
modes          600 and 664, unchanged
symlink        current -> app.yml, intact
nested dir     sub/deep.txt, intact
contents       identical
named volume   restored, mode 640 intact
```

The outer archive is unpacked by the new extractor, so this exercises it on a file homebutler wrote rather than one assembled by a test.

PR #117, opened 2026-08-30
refactor: consolidate port parser tests into internal/ports and un-export parsers

### LINKED ISSUE #38: Consolidate port parser tests into internal/ports and un-export the parsers

lenny-ts:
Fixture tests in `internal/inventory` existed there by mistake — the parsing
lives in `internal/ports`, not inventory. That split forced `ParseLinuxOutput`
and `ParseDarwinOutput` to be exported purely so another package's tests could
reach them.

Move the fixture tests and testdata into `internal/ports`, merge them with the
existing inline tests, and revert both parsers to unexported names.

- Move `testdata/{ss-linux,lsof-macos}.txt` to `internal/ports/testdata/`
- Add fixture-driven tests using `findPortByNumber` lookup instead of
  index-based `ports[0]`/`ports[2]` coupling
- Drop overlapping inline-string tests, keep edge cases fixtures don't cover
  (empty input, header-only, dedup, IPv6, multiple processes, no process info)
- Delete `internal/inventory/portparse_fixture_test.go` and its testdata

Parser tests now live only in `internal/ports`. Coverage is not reduced
relative to the union of the two previous test files. `go test ./...` passes.

close #38


Higangssh (approved):
Does what #38 asked for, and I checked the claims rather than taking them.

`ParseLinuxOutput` and `ParseDarwinOutput` are gone from the package surface with nothing left calling the exported names. The testdata is a real move — git records it as a rename, not a copy — and the helpers came with it rather than being duplicated. Coverage is unchanged at 67.6% for `ports` and 91.9% for `inventory`.

The part I want to call out is the one the issue asked for almost in passing. Dropping `TestParseLinuxOutput` and `TestParseDarwinOutput` in favour of the fixture tests looks like losing two tests, and it isn't: they indexed into `ports[0]` and `ports[2]`, so they were asserting parser ordering as much as parser output, and `findPortByNumber` asserts the same fields without that coupling. Process, PID, address and the 127.0.0.1 bind are all still covered. That is a better test than what it replaced.

Thanks for picking up a cleanup issue — the split this undoes was my mistake in #29, and it is the kind of thing that stays wrong for a long time because nothing breaks while it is.

Issue #106, opened 2026-08-30
dashboard: show Proxmox freshness and stale-data state


gsaraiva2109:
## Problem

The Proxmox dashboard can show the last successful snapshot without showing whether it is current. If a later refresh fails, stale data can look healthy. A valid empty response must also remain distinct from data that is unavailable or ACL-filtered.

## Scope

Add a compact freshness and availability state to the existing Proxmox dashboard panel:

- last successful refresh time for each endpoint
- current refresh status: current, stale, partially readable, unavailable, or not configured
- collector-level reason when relevant: TLS verification, authentication, ACL/authorization, or transport failure
- retain and clearly label the last successful data while a refresh failure is active

The UI should say what it knows. For example, `Updated 20 seconds ago` is distinct from `Last successful update 15 minutes ago: TLS validation failed`. Empty but readable data stays an empty state, not an error.

## Boundaries

- Read-only dashboard behavior only. No action controls, retries of state-changing calls, credential display, polling loop, scheduler, or notifications.
- Reuse the server-side partial-result and error classification already returned by the Proxmox status endpoint. Do not add a second Proxmox request path in the browser.
- Do not treat stale cache as live truth. Show its age and degraded status.

## Done when

- A successful refresh displays its timestamp.
- Partial collector failure preserves readable data and names the failed collector.
- A later failure labels retained data as stale with a safe error class.
- Empty readable data, ACL-filtered data, and unavailable data have distinct accessible UI states.
- Frontend tests cover each state without exposing endpoint credentials.

Related: #79 for the Proxmox dashboard, #105 for endpoint diagnostics, and #104 for watcher error classification.

Higangssh:
Accepted. The distinction this is built on — that stale data can look healthy — is the same one behind `failed_collectors` in #74, and the dashboard is currently the place where it is lost: `proxmox status` reports which collectors failed, and the panel renders the last good snapshot without saying when it was good.

Depends on #111, filed off this issue and #104 and #105. The reason to reuse a server-side classification rather than let the frontend interpret is exactly what you wrote in the boundaries, and today there is nothing to reuse — the classification does not exist on the server side either.

One thing to add to the scope: `Updated 20 seconds ago` has to be honest about which clock. If refresh time is stamped in the browser it drifts from the endpoint's own view, and a dashboard left open overnight will confidently report a freshness it cannot know. The timestamp should come from the same response that carries the data.

Issue #105, opened 2026-08-30
doctor: add read-only Proxmox endpoint preflight


gsaraiva2109:
## Problem

A Proxmox endpoint can be syntactically configured yet still fail through TLS validation, a missing or unreadable token file, an invalid token, privilege-separated ACLs, or partial collector permissions. Today the user has to run several Proxmox commands and interpret HTTP errors themselves.

The general `doctor` work in #99 already covers warning about `insecure: true`. This issue covers live read-only endpoint diagnosis.

## Scope

Add a Proxmox section to `homebutler doctor`, one result per configured endpoint, using the existing read-only client and existing status collectors. It reports:

- configuration presence and whether the token file can be opened, without printing its content
- TLS trust result, including fingerprint or CA mismatch as a distinct failure
- API authentication result
- readable status collectors: version, cluster status, and resources
- ACL-filtered collectors separately from unavailable endpoints
- partial visibility without replacing valid zero resources with a failure

Each finding says what the operator can safely inspect or correct, for example the configured certificate trust method, token ID, token-file ownership, or the read-only `PVEAuditor` ACL applied to both the API user and privilege-separated token. It must never recommend `Administrator` merely to make a check pass.

## Boundaries

- GET requests only, with no guest actions, no SSH fan-out, no mutations, and no token value in output, logs, JSON, or MCP responses.
- Reuse the configured TLS and timeout behavior. Do not add a second credential path or silently bypass certificate validation.
- Do not collapse empty, ACL-filtered, authentication-failed, TLS-failed, and unreachable states into one generic unavailable result.
- Do not add a scheduler, watcher, notification, or automatic repair. #104 can consume the same error classification later.

## Done when

- `doctor` identifies a valid fully readable endpoint, a partially readable endpoint, and an endpoint with no readable collectors.
- TLS, token-file, authentication, authorization, and transport failures are distinguishable and secret-safe in human and JSON output.
- A response that is valid but empty remains visibly different from a collector that could not be read.
- Focused fixtures cover fingerprint mismatch, invalid token, `Sys.Audit` denial, empty resources, and partial collector success.

Related: #32 for the Proxmox surface, #99 for general doctor findings, and #104 for read-only Proxmox monitoring.

Higangssh:
Overlaps #99, which adds five local checks to `doctor`. Worth building on the same footing.

[...]

Issue #104, opened 2026-08-30
watch: add read-only Proxmox endpoint and guest-state incidents


gsaraiva2109:
## Problem

Proxmox data is available through explicit read-only commands, but HomeButler cannot notify an operator when a configured endpoint becomes unavailable or when a guest that is meant to run stops. A generic watcher or a second scheduler would duplicate the work in #97 and #80.

## Scope

Add Proxmox as a read-only source to the unified `watch start` runtime from #97, using the lifecycle/service work from #80. It must not introduce another long-running command, scheduler, or service unit.

A configuration entry names each expected-running guest exactly by `endpoint`, `node`, `type` (`qemu` or `lxc`), and `vmid`. Guests not declared there are observational only: a deliberately stopped guest must not produce an incident.

Report state transitions only:

- endpoint unavailable and recovered
- endpoint authenticated but ACL-filtered and recovered
- configured guest not running and recovered

Failures must retain their class: TLS verification, authentication/token expiry, authorization/ACL, transport/timeout, and valid empty result are distinct states. Alerts use the one notifier and incident store from #97.

## Boundaries

- API GETs only; no guest actions, remediation, POSTs, SSH fan-out, or token output.
- No automatic retry of a state-changing action because this issue performs none.
- No task-history polling, storage thresholds, resource thresholds, discovery/baselining, bulk operations, or AI decisions. Those need separate evidence and designs.
- Preserve partial results. One failed collector must not erase readable node or guest data.

## Done when

- A configured expected-running QEMU or LXC guest emits one incident on transition to non-running and one recovery incident when running again.
- A transient endpoint failure produces one unavailable incident and a recovery, without treating it as an empty cluster.
- TLS, auth, ACL, and transport failures are distinguishable in human output, JSON, and incident records without credential leakage.
- Unconfigured and intentionally stopped guests do not alert.
- Focused fixtures cover a standalone node, partial API visibility, and each failure class.

Depends on #97 for the unified monitoring loop and should use the lifecycle installed by #80. Builds on the read-only Proxmox surface in #32.

Higangssh:
Accepted, and the boundaries are the right ones — particularly that an undeclared guest is observational only. A watcher that alerts on every stopped guest would be turned off within a day, and `watch` already has that lesson baked in: it tracks declared targets, not everything it can see.

Two things worth settling before this is written.

[...]

PR #120, opened 2026-08-31
feat(proxmox): classify endpoint failures

### LINKED ISSUE #111: proxmox: API failures are flat strings, so nothing downstream can tell them apart
## Problem

Every failure out of `internal/proxmox` is a formatted string. `client.go:372`:

```go
return fmt.Errorf("proxmox request %s: %s: %s", path, resp.Status, message)
```

The HTTP status is rendered into the message and then gone. A caller that wants to know whether it got a 401 or a 403 has to match on text, and nothing does, so today every failure reaching the CLI, the dashboard and MCP is the same failure.

That is fine while the only consumer is `proxmox status`, which reports `failed_collectors` and stops there. It stops being fine immediately: #104, #105 and #106 each independently require that TLS verification, authentication, authorization/ACL, transport, and a valid-but-empty result stay distinguishable. Three issues asking for the same classification, none of which can produce it.

Built separately they would produce three of them, in three packages, disagreeing at the edges — which is worse than none, because the disagreement would be invisible until an operator got a wrong diagnosis out of `doctor` for a condition the dashboard was rendering correctly.

## Scope

A typed failure classification in `internal/proxmox`, produced where the failure actually happens and carried outward:

- **TLS** — fingerprint mismatch, CA mismatch, and no-certificate are already distinguished at `client.go:119-156` and immediately flattened
- **Authentication** — 401, and a token file that cannot be read, which fails earlier but lands the operator in the same place
- **Authorization** — 403, which for a privilege-separated token means the intersection of the user's ACL and the token's, and is the failure most likely to be misread as an authentication problem
- **Transport** — dial, timeout, DNS
- **Empty but readable** — not a failure. `failed_collectors` exists because an empty response was being reported as a real reading; the classification has to preserve that distinction rather than re-flatten it

Callers ask the class; they do not parse the message. The message stays human-facing and keeps saying what to do about it.

## Boundaries

No behaviour change for anything that exists today. `proxmox status` keeps its current output and its `failed_collectors` semantics from #74. This is the shared vocabulary the three issues above are each about to invent.

## Done when

- Each class is reachable from a fixture, including a privilege-separated 403 distinct from a 401
- A valid empty result is not a failure class
- No classification carries a token value, and the token ID is the most that appears in output

Prerequisite for #104, #105 and #106. Related: #107, which needs read and action credentials to fail independently and be reported as such.

gsaraiva2109:
## What does this PR do?

Adds typed Proxmox failure classes for TLS, authentication, authorization, and transport failures.

The shared class survives error wrapping without changing existing human-facing messages or DefaultView failed collector behavior. Token file read failures are classified as authentication failures.

Closes #111

## Checklist

- [x] `gofmt -w .`
- [x] `go build ./...`
- [x] `go test ./...`
- [x] README update not needed for this internal API change

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/Higangssh/homebutler/internal/backup"
 	"github.com/Higangssh/homebutler/internal/notify"
+	"github.com/Higangssh/homebutler/internal/proxmox"
 	"github.com/Higangssh/homebutler/internal/watch"
 	"gopkg.in/yaml.v3"
 )
@@ -209,12 +210,12 @@ func (p ProxmoxConfig) TokenValue() (string, error) {
 
 	data, err := os.ReadFile(p.TokenFilePath())
 	if err != nil {
-		return "", fmt.Errorf("read Proxmox token file %s: %w", p.TokenFile, err)
+		return "", proxmox.WithFailureClass(proxmox.FailureAuthentication, fmt.Errorf("read Proxmox token file %s: %w", p.TokenFile, err))
 	}
 	if token := strings.TrimSpace(string(data)); token != "" {
 		return token, nil
 	}
-	return "", fmt.Errorf("proxmox token file %s is empty", p.TokenFile)
+	return "", proxmox.WithFailureClass(proxmox.FailureAuthentication, fmt.Errorf("proxmox token file %s is empty", p.TokenFile))
 }
 
 type WakeTarget struct {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -3,8 +3,10 @@ package config
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/Higangssh/homebutler/internal/proxmox"
 	"gopkg.in/yaml.v3"
 )
 
@@ -251,14 +253,18 @@ func TestProxmoxConfigTokenValue(t *testing.T) {
 		t.Error("TokenValue() should fail without a token source")
 	}
 
-	if _, err := (ProxmoxConfig{TokenFile: filepath.Join(dir, "missing")}).TokenValue(); err == nil {
+	if _, err := (ProxmoxConfig{TokenFile: filepath.Join(dir, "missing"), Token: "inline-secret"}).TokenValue(); err == nil {
 		t.Error("TokenValue() should fail for a missing token file")
+	} else if proxmox.Classify(err) != proxmox.FailureAuthentication || strings.Contains(err.Error(), "inline-secret") {
+		t.Errorf("TokenValue() error = %v, want authentication error without inline token", err)
 	}
 	if err := os.WriteFile(filepath.Join(dir, "empty"), []byte(" \n"), 0600); err != nil {
 		t.Fatalf("write empty token: %v", err)
 	}
-	if _, err := (ProxmoxConfig{TokenFile: filepath.Join(dir, "empty")}).TokenValue(); err == nil {
+	if _, err := (ProxmoxConfig{TokenFile: filepath.Join(dir, "empty"), Token: "inline-secret"}).TokenValue(); err == nil {
 		t.Error("TokenValue() should fail for an empty token file")
+	} else if proxmox.Classify(err) != proxmox.FailureAuthentication || strings.Contains(err.Error(), "inline-secret") {
+		t.Errorf("TokenValue() error = %v, want authentication error without inline token", err)
 	}
 }
 
--- a/internal/proxmox/client.go
+++ b/internal/proxmox/client.go
@@ -153,14 +153,14 @@ func tlsConfig(options Options) (*tls.Config, error) {
 		config.InsecureSkipVerify = true // #nosec G402 -- verified by the callback below.
 		config.VerifyPeerCertificate = func(rawCerts [][]byte, _ [][]*x509.Certificate) error {
 			if len(rawCerts) == 0 {
-				return fmt.Errorf("proxmox TLS peer sent no certificate")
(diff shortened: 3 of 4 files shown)


Higangssh (changes requested):
This is the right shape. The class rides on a wrapper with `Unwrap`, so `errors.As` finds it however far the error is wrapped, and `Error()` delegates — the human-facing messages are byte-identical, and the `[REDACTED]` token masking is untouched. `WithFailureClass` declining to overwrite an existing class means the source-most classification wins, which is what you want when an error passes several layers. The 401/403 subtests asserting the token does not appear in the message is the kind of check I would have asked for.

One change I would like before merging, and one gap.

**The `strings.Contains(err.Error(), "proxmox TLS ")` branch should not be needed.**

This is the one thing in the PR that reads a message to make a decision, in the PR whose purpose is to let everything downstream stop doing that. It is there because the fingerprint and no-certificate errors come out of the `VerifyPeerCertificate` callback, and Go does not wrap a callback's error in `*tls.CertificateVerificationError` — so `isTLSFailure` genuinely does not see them. I confirmed that: for a fingerprint mismatch, `isTLSFailure(err)` is `false`.

But classifying at the point the failure happens does work. Wrapping the two callback returns:

```go
return WithFailureClass(FailureTLS, fmt.Errorf("proxmox TLS peer sent no certificate"))
return WithFailureClass(FailureTLS, fmt.Errorf("proxmox TLS certificate fingerprint mismatch"))
```

lets the `strings.Contains` clause go, and `Classify` still returns `tls` through the handshake and `*url.Error`. I ran it against a real `httptest.NewTLSServer` with a mismatched fingerprint — same class, same message, and `internal/proxmox`, `internal/config`, `cmd` and `internal/mcp` all still pass.

It is also the pattern the rest of the PR already uses: `New` classifies where the error is made rather than where it is caught.

**The gap: nothing covers a fingerprint mismatch.**

`TLS handshake` uses a server with an untrusted cert, which is an `x509.UnknownAuthorityError` and goes through the typed check. The fingerprint path never runs. I deleted the `strings.Contains` clause outright, with no other change, and every test still passed — so that branch has no coverage at all.

That matters more than an ordinary uncovered line, because certificate pinning is homebutler's own TLS feature rather than something the standard library hands us, and a mismatch is the failure an operator most needs named correctly. Worth a subtest that sets `Fingerprint` to something that will not match and asserts `FailureTLS`.

**Minor, take it or leave it:** `New` wraps everything out of `tlsConfig` as `FailureTLS`, including `read Proxmox CA file` and `proxmox fingerprint must be a SHA-256 certificate fingerprint`. Those are configuration problems rather than verification failures — an unreadable CA file is closer to the token-file case this PR classifies as authentication. It points the operator at the right area either way, so I am not asking for a change; just noting it in case `doctor` in #105 wants to tell those apart.

Thanks for taking the prerequisite rather than one of the three issues that are more visible. #104, #105 and #106 each independently asked for these five classes, and doing this first is what stops them inventing three of them.

gsaraiva2109:
Pushed eb0ab9f. Both callback returns in `tlsConfig` now classify at the source with `WithFailureClass(FailureTLS, ...)`, and the `strings.Contains(err.Error(), "proxmox TLS ")` clause in `request` is gone; `Classify(err) == FailureTLS || isTLSFailure(err)` is enough now that the callback errors carry their own class through the wrap chain.

`TestTLSModes/fingerprint_mismatch` already asserts `Classify(err) == FailureTLS` for a mismatched fingerprint, so that path stays covered with the string match removed. `go test ./...`, `go vet ./...`, and `golangci-lint run` are all clean.

Left the CA file and fingerprint format errors classified as `FailureTLS` for now, as suggested, since that is a separate question for #105's doctor command.

gsaraiva2109:
Pushed f0cbc43: went ahead and split the minor point too. CA file read/empty and fingerprint-format errors now classify as FailureAuthentication (same bucket as the token-file case), leaving FailureTLS for actual verification failures (mismatch, no cert, handshake). Renamed the old "TLS setup" subtest to "CA file unreadable" for its new expectation and added a "fingerprint format" subtest. All five checks still clean.

Higangssh (approved):
**A correction first.** I said nothing covered a fingerprint mismatch and asked you for a subtest. That was wrong. `TestTLSModes/fingerprint_mismatch` was already there in `c8eff80` and already asserted the class. I re-ran the experiment I claimed to have run — removing the `strings.Contains` clause from `c8eff80`, changing nothing else — and it fails that subtest, exactly as you said it would. Thanks for saying so instead of adding a duplicate test to satisfy the review.

The classification move itself is right, and the diff is smaller than the problem it solves. Both callback returns carry `FailureTLS` from the point the failure happens, `request` is back to two typed checks, and the string match is gone from the one PR whose purpose was to stop reading messages to make decisions. `strings` is still earning its import through `parseFingerprint`, so nothing was left dangling.

On the split you went ahead with anyway: the CA file half is right. An unreadable or empty trust anchor is a credential problem in the same way the token file is, and renaming the subtest to `CA file unreadable` makes it say what it now asserts.

**One observation, and nothing to do about it here.** `proxmox fingerprint must be a SHA-256 certificate fingerprint` is a typo in a YAML value, and `authentication` is the closest of the four classes rather than an accurate one — there is no configuration class, and an error out of `New` is already known to be a construction failure by where it came from. It costs nothing today: `classifiedError` carries no JSON tag and `Classify` is read only inside `internal/proxmox` and `internal/config`, so nothing acts on it yet. Worth revisiting when #105 decides what `doctor` says for each class, since that is the first consumer that would tell an operator to check their credentials over a mistyped fingerprint.

CI is green, and `gofmt`, `go vet` and the `proxmox`, `config`, `cmd` and `mcp` tests pass locally on `f0cbc43`.

Thanks — for taking the prerequisite in the first place, and for taking the minor point after I had explicitly said not to bother. #104, #105 and #106 all start from one vocabulary now instead of inventing three.

PR #121, opened 2026-09-01
feat: show actionable errors before details

### LINKED ISSUE #45: Add a global --verbose flag so error output can lead with the fix

lenny-ts:
Permission errors now lead with the actionable fix instead of the raw filesystem error.

The new global `--verbose/-v` flag preserves the full underlying error when more detail is needed. Structured hint errors are used consistently across backup, restore, install, and upgrade paths.

`--verbose` is preserved when forwarding commands to remote servers, while `--json` output remains unchanged.

Verification:
- `go test ./...`
- `go vet ./...`
- `golangci-lint run`
- `go build ./...`

close #45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ All notable changes to this project will be documented in this file.
 
 - add the global `--verbose` / `-v` flag for detailed error output. Human-facing permission errors now include the failed operation and a command that preserves the requested backup options; remote commands receive the flag too
 
+### ⚠️ Behavior changes
+
+- **Human-facing permission errors no longer include the underlying cause by default.** They lead with the failed operation and the command to rerun; use `--verbose` to see the underlying error and hint together
+
 ## [0.24.0](https://github.com/Higangssh/homebutler/compare/v0.23.1...v0.24.0) - 2026-08-31
 
 **The largest files homebutler writes were the only ones nothing ever deleted, and the safety of restoring them belonged to a program homebutler merely invoked.** `backup` wrote an archive on every run with no retention of any kind — fine while a person types the command, and the first thing to hurt whoever puts it in cron. Restore, meanwhile, held its destination only because GNU tar and bsdtar decline dangerous member names on their own; nothing in this repository asserted it and no test covered it. Separately, the MCP server had been answering every handshake with the first protocol revision ever published.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+**Permission failures now lead with the fix while keeping the underlying error available with `--verbose`.**
+
+### ✨ Features
+
+- add the global `--verbose` / `-v` flag for detailed error output. Human-facing permission errors now include the failed operation and a command that preserves the requested backup options; remote commands receive the flag too
+
 ## [0.24.0](https://github.com/Higangssh/homebutler/compare/v0.23.1...v0.24.0) - 2026-08-31
 
 **The largest files homebutler writes were the only ones nothing ever deleted, and the safety of restoring them belonged to a program homebutler merely invoked.** `backup` wrote an archive on every run with no retention of any kind — fine while a person types the command, and the first thing to hurt whoever puts it in cron. Restore, meanwhile, held its destination only because GNU tar and bsdtar decline dangerous member names on their own; nothing in this repository asserted it and no test covered it. Separately, the MCP server had been answering every handshake with the first protocol revision ever published.
--- a/README.md
+++ b/README.md
@@ -597,6 +597,7 @@ Commands:
 
 Flags:
   --json              JSON output (default: human-readable)
+  --verbose, -v       Show detailed error information
   --server <name>     Run on a specific remote server
   --all               Run on all configured servers in parallel
   --port <number>     Port for serve command (default: 8080)
@@ -670,6 +671,7 @@ Commands:
 
 Flags:
   --json              JSON output (default: human-readable)
+  --verbose, -v       Show detailed error information
   --server <name>     Run on a specific remote server
   --all               Run on all configured servers in parallel
   --port <number>     Port for serve command (default: 8080)
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -17,6 +18,7 @@ import (
 	"github.com/Higangssh/homebutler/internal/ports"
 	"github.com/Higangssh/homebutler/internal/remote"
 	"github.com/Higangssh/homebutler/internal/system"
+	"github.com/Higangssh/homebutler/internal/util"
 	"github.com/Higangssh/homebutler/internal/wake"
 )
 
@@ -28,10 +30,11 @@ var (
 
 // Global flags
 var (
-	jsonOutput bool
-	serverName string
-	allServers bool
-	cfgPath    string
+	jsonOutput    bool
+	verboseOutput bool
+	serverName    string
+	allServers    bool
+	cfgPath       string
 )
 
 // cfg holds the loaded config (set in root PersistentPreRun)
@@ -111,6 +114,13 @@ func output(data any, jsonOut bool) error {
 		return enc.Encode(data)
 	}
 	return nil
+}
+
+func formatCommandError(err error) error {
+	if err == nil || jsonOutput {
+		return err
+	}
+	return errors.New(util.FormatError(err, verboseOutput))
 }
 
 // listServerNames returns a formatted list of server names from config.
@@ -231,7 +241,7 @@ func runAllServers(c *config.Config, args []string, jsonOut bool) error {
 			if server.Local {
 				out, err := runLocalCommand(remoteArgs)
 				if err != nil {
-					result.Error = err.Error()
+					result.Error = util.FormatError(err, verboseOutput || jsonOut)
 				} else {
 					result.Data = json.RawMessage(out)
 				}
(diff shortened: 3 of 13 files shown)


Higangssh (changes requested):
The design is the one #45 argued for, and it is the harder of the two to get right: a typed error carrying `{cause, hint}`, formatted at the display layer, so nothing in `internal/` learns about verbosity and there is no global to reset between tests. `SilenceErrors: true` keeps `main.go` the only thing that prints, so there is no double output to untangle. Keeping `--verbose` out of the `filterFlags` strip list is the decision #45 asked for, and the test pins it so a later edit cannot quietly reverse it.

**One change before merging.** `FormatError` prints a fixed sentence — `"permission denied — rerun with: " + Hint` — and drops the cause. Both halves of that give me trouble.

"permission denied" is asserted for any `HintError`, while `NewHintError` is exported and documented generically. `util.IsPermissionError` also matches EPERM, which is "operation not permitted", and `install.go:489` already carries a docker-not-running hint that would naturally move onto this type. I built your branch and ran one through:

```
permission denied — rerun with: sudo systemctl start docker
```

The other half is that the *where* leaves with the syscall noise. #45 wanted `rename /usr/local/bin/homebutler …: permission denied` to stop leading; it did not ask for the path homebutler was writing to. `restore` loses `m.Source` — with `--allow-bind` that is the one thing an operator needs to see — and the five `install` hints collapse into one sentence that can no longer tell "could not create the data dir" from "could not record the install".

Giving `HintError` a field for the actionable half fixes both. I ran this on your branch:

```go
type HintError struct {
	Op    string // what failed, including the path it failed on
	Cause error
	Hint  string
}

// short:   fmt.Sprintf("%s — rerun with: %s", e.Op, e.Hint)
// verbose: unchanged
```

```
cannot create backup dir /mnt/nas/backups — rerun with: sudo homebutler backup
cannot create bind mount dir /srv/appdata — rerun with: sudo homebutler restore /backups/backup_20260901_143012.812.tar.gz
docker daemon is not running — rerun with: sudo systemctl start docker
```

Another shape is fine. What matters is that the sentence comes from the failure rather than being asserted over it.

**Three hints that were survivable as trailing advice are not survivable as the whole message.** Each is now the only thing printed, so each has to work when someone types it.

- `internal/backup/restore.go:313` — `archivePath` here is `mountArchivePath(m.Name, volDir)`, a path inside the `os.MkdirTemp` directory `Restore` deletes on the way out (`restore.go:126-131`). It has always named a file that is gone by the time anyone reads it; it just used to trail a real error. Now `sudo homebutler restore /var/folders/…/homebutler-restore-123/volumes/appdata.tar.gz` is the whole output, and it answers `archive not found`. It wants the archive the operator named. (Read rather than run — I had no way to exercise the bind path here.)
- `internal/install/install.go:39` — `saveInstalled` runs after `docker compose up -d` has already succeeded (`install.go:705-711`). "rerun with: sudo homebutler install jellyfin", standing alone, tells someone to reinstall an app that is up and running when all that failed was the registry write.
- `internal/backup/backup.go:103,109` — the hint is a bare `sudo homebutler backup` while `--to` and `--service` came from the command line (`cmd/backup.go:29-34`). After `homebutler backup --to /mnt/nas --service jellyfin` fails, obeying the printed command backs up every service to the default directory, and there is no cause line left to reveal the substitution.

**Minor, and the last one is mine rather than yours.**

- `cmd/helpers.go:244` is inert: `runLocalCommand` dispatches only `status`, `alerts`, `docker list`, `ports` and `doctor`, none of which construct a `HintError` — and the remote branch eight lines below still calls `err.Error()`. If one ever does reach here, `--all` renders the local server short and every remote server long in the same block.
- The README lists the global flags in two places and neither gains `--verbose`, and there is no CHANGELOG entry. Someone who meets the new one-line error has no documented way to find out `-v` exists.
- `cmd/ports.go:31` staying in the old shape is right, I think — it is a warning printed after output that worked, not an error, so there is no ordering problem there. Worth saying out loud since #45 listed it among the nine.
- Forwarding `--verbose` to a remote is what I asked for in #45, and I was not thinking about version skew when I wrote it. A 0.24.0 binary answers `unknown flag: --verbose`, and `internal/remote/ssh.go:41` leads that with "Check if homebutler is installed on the remote server" — the real output is printed underneath, but the advice points elsewhere. Not yours to solve here; I will carry it as a release note.

CI is green, and `gofmt`, `go vet` and the touched packages' tests pass locally on the branch.

Thanks for taking this one. It is the second time you have picked the piece that has to exist before anything else can — #117 moved the parser tests to where parsing actually lives so the exports could go, and this gives every future hint somewhere to live instead of another format string.

lenny-ts:
Thanks for the detailed review. I updated HintError to carry Op, Cause, and Hint, so short output now preserves the failed operation and relevant path without asserting a fixed “permission denied” sentence. Verbose output still includes the underlying cause and hint.

I also fixed the actionable commands:
- restore now uses the archive path originally supplied by the operator
- backup preserves --to and --service in the suggested command
- install registry failures suggest fixing ownership instead of rerunning an already completed install
- --all formats both local and remote errors consistently

The global flag is documented in both README flag sections and CHANGELOG, and --verbose remains forwarded to remote commands.

Higangssh (approved):
That is all four, and the shape you landed on is better than the one I sketched. `Op` carrying the path means `Error()` reads as a sentence in verbose output too, and putting `backupHint` and `fixPermissionsHint` next to their call sites keeps command construction where the flags are known instead of pushing it into `util`.

The three that were the whole point:

- `restoreMount` takes the operator's archive now and `Restore` passes it down, so the printed command names a file that exists. The temp path is gone from the hint entirely.
- `backupHint` keeps `--to` and `--service` through `ShellQuote`, and it emits `--to` even when the destination came from config rather than the flag. That is the right call — the printed command has to do the same thing whether or not the operator typed the destination.
- `saveInstalled` no longer proposes reinstalling something that is already running, and ownership is the correct diagnosis for that failure.

`runAllServers` formats both arms the same way now, so `--all` cannot render local and remote differently in one block.

`gofmt`, `go vet` and the `util`, `backup`, `install`, `cmd` and `remote` tests pass locally on `ab5cfb0`.

**One line I would still change, and it is the last thing.** `fixPermissionsHint(BaseDir())` in the `MkdirAll` branch names a directory that does not exist yet — that branch runs precisely because `~/.homebutler/apps` could not be created:

```
$ chown -R $(id -u):$(id -g) ~/.homebutler/apps
chown: /Users/…/.homebutler/apps: No such file or directory
```

The parent is the thing whose ownership is wrong, so `filepath.Dir(BaseDir())` in that branch. The two `WriteFile` branches are right as they stand — the file or its directory exists by the time they run.

**And one for the changelog rather than the code.** The entry describes the feature but not the behaviour change underneath it: default error output no longer contains the underlying cause, so anything reading homebutler's stderr sees a different string after this. `⚠️ Behavior changes` is the section people read before upgrading. Add a line there, or leave the changelog to me and I will carry it into the release notes.

Approving so this goes in as soon as that one line lands. Thanks for turning it around this fast, and for taking the shape further than the sketch rather than transcribing it.

lenny-ts:
The final review point is addressed.

The MkdirAll registry error now points fixPermissionsHint at filepath.Dir(BaseDir()), which is the existing directory whose ownership can actually be corrected. The WriteFile cases still target the existing registry file or directory.

I also added the default error-output change to the CHANGELOG Behavior changes section: permission errors no longer include the underlying cause by default, and --verbose restores the detailed output.

PR #123, opened 2026-09-01
doctor: add read only Proxmox endpoint preflight

### LINKED ISSUE #105: doctor: add read-only Proxmox endpoint preflight

gsaraiva2109:
## Summary

doctor now checks each configured Proxmox endpoint using read only requests (version, cluster status, resources) and reports one finding per endpoint. TLS, authentication, authorization, and transport failures stay distinguishable instead of collapsing into a single unavailable result, and a valid but empty response is never treated as a failure.

Authorization findings point at the read only PVEAuditor role and never suggest widening a token to Administrator to make a check pass.

Builds on the failure classification merged in #120 (#111).

## Test plan

- [x] gofmt -w .
- [x] go vet ./...
- [x] golangci-lint run
- [x] go test ./...
- [x] go build ./...

Closes #105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### ✨ Features
+
+- diagnose each configured Proxmox endpoint in `doctor` with read-only requests (#105). `doctor` said nothing about Proxmox before this; now every endpoint gets one finding, using the same `DefaultView` call `proxmox status` makes, so the two never disagree about what a token can reach. TLS, authentication, authorization, and transport failures stay distinguishable, and the action `doctor` prints never suggests widening a token to Administrator to make a check pass — it names the read-only PVEAuditor role instead
+
+### ⚠️ Behavior changes
+
+- **`doctor` now makes outbound network calls, one per configured Proxmox endpoint.** A host that is slow to answer or unreachable makes `doctor` take noticeably longer, and `--strict` cron jobs now exit non-zero for a Proxmox host that is merely rebooting or firewalled, not only for problems on the local machine
+
 ## [0.25.0](https://github.com/Higangssh/homebutler/compare/v0.24.0...v0.25.0) - 2026-09-02
 
 **Every permission failure homebutler can hit printed the operator's next command last.** `backup`, `restore`, `install` and `upgrade` all led with the raw syscall error and put the one line that mattered underneath it — and the detail could not simply be dropped, because there was no way to ask for it back. This release inverts the order and adds the flag that makes dropping it safe.
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ homebutler report --json
 
 - **Install apps** — deploy Uptime Kuma, Jellyfin, Pi-hole, Gitea, Portainer, and more with one command
 - **Map your server** — see containers, exposed ports, system ports, and service topology
-- **Run a doctor check** — diagnose resource pressure, stopped containers, public ports, backup hygiene, notifications, and report baseline readiness
+- **Run a doctor check** — diagnose resource pressure, stopped containers, public ports, backup hygiene, notifications, report baseline readiness, and configured Proxmox endpoint reachability
 - **Catch crashes** — save logs before/after Docker, systemd, or PM2 restarts and detect flapping loops
 - **Verify backups** — boot backups in isolated containers before you trust them
 - **See a Proxmox cluster** — nodes, QEMU and LXC guests, storage, and task status, with power actions that name their target explicitly
@@ -134,7 +134,7 @@ homebutler doctor --json            # automation / MCP friendly
   <img src="assets/doctor-card.svg" alt="homebutler doctor reporting a full disk, a stopped container, and a missing report baseline, each with the command to run next" width="700">
 </p>
 
-`doctor` is a read-only preflight for the problems homelab users usually discover too late: high disk or memory usage, stopped containers, public bind ports, stale or missing backups, missing notifications, and whether `report` has a baseline for change detection. Every finding names the next command to run, so `--strict` makes it usable from cron or CI.
+`doctor` is a read-only preflight for the problems homelab users usually discover too late: high disk or memory usage, stopped containers, public bind ports, stale or missing backups, missing notifications, whether `report` has a baseline for change detection, and whether each configured Proxmox endpoint is reachable with the token it has. Every finding names the next command to run, so `--strict` makes it usable from cron or CI — including a Proxmox host that is unreachable or rebooting.
 
 ### 🗂 Config Validation
 
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -17,7 +17,8 @@ func newDoctorCmd() *cobra.Command {
 		Short: "Diagnose homelab health, exposure, backups, and readiness",
 		Long: `Run a read-only diagnosis for the things that usually hurt self-hosted servers:
 resource pressure, stopped containers, public bind ports, backup hygiene,
-notification readiness, and report baseline status.`,
+notification readiness, report baseline status, and configured Proxmox
+endpoint reachability.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := loadConfig(); err != nil {
 				return err
--- a/docs/mcp-server.md
+++ b/docs/mcp-server.md
@@ -78,7 +78,7 @@ Restart your AI client — homebutler tools will appear automatically.
 |--------------------------|------------------------------------------------------------------------------------------------------------------------|
 | `system_status`          | CPU, memory, disk, uptime                                                                                              |
 | `report`                 | Butler-style health report with snapshot comparison and suggested actions                                              |
-| `doctor`                 | Read-only diagnosis of resource pressure, stopped containers, public ports, backup hygiene, and notification readiness |
+| `doctor`                 | Read-only diagnosis of resource pressure, stopped containers, public ports, backup hygiene, notification readiness, and configured Proxmox endpoint reachability |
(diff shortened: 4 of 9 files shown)


Higangssh (changes requested):
This does what #105 asked, and the part I would have checked first is the part you got right: `proxmoxFailureAction` never suggests widening a token to make a check pass, it names PVEAuditor and the user/token intersection instead, and an endpoint that authenticates and returns an empty list is a Pass rather than a failure. Injecting `ProxmoxOpenFn` means the tests exercise the classification rather than a network, and `openProxmoxEndpoint` resolves the credential exactly the way the commands do.

Three things before this merges, and the first is mine more than yours.

**1. The class is wrong for CA and fingerprint problems, and this is the PR where that stops being theoretical.**

I closed the #120 review by noting that `New` wraps everything out of `tlsConfig` as `FailureAuthentication`, including `read Proxmox CA file` and the fingerprint format error, and said I was not asking for a change — only flagging it in case `doctor` in #105 wanted to tell them apart. This is #105, and it wants to. A missing `ca_file` produces:

```
detail: read Proxmox CA file: open /does/not/exist.pem: no such file or directory
action: Check the token ID, the token value, and the token file's ownership and
        permissions; the token may be missing, unreadable, or revoked.
```

The token is fine. And `WithFailureClass` will not overwrite an existing class — the property that makes the source-most classification win everywhere else — so `New`'s own `WithFailureClass(FailureTLS, err)` cannot correct it from outside either.

The fix belongs in `internal/proxmox/client.go`, not in a branch here: either a class for configuration material, or leave construction errors unclassified so the caller reads them as what they are. Teaching `doctor` to distrust the class would undo the thing #111 was for.

**2. `doctor` now blocks on the network with no deadline and no way to cancel.**

`checkProxmoxEndpoint` makes three sequential requests per endpoint under `context.Background()`, and `Run` takes no context. Each is bounded only by the endpoint's own timeout — 10s by default, and `internal/config/validate.go` accepts any positive duration, so `timeout: 5m` is legal. Three endpoints behind a firewall that drops packets turn a command that was local and instant into a 90-second hang, or a 45-minute one.

That lands in unattended jobs first: README:137 advertises `--strict` for cron and CI. The same code runs behind the MCP `doctor` tool, where the client's own deadline can expire with three requests still in flight and nothing able to cancel them.

At minimum, stop after the first `FailureTransport` — the second and third calls re-pay the same dial timeout for no new information. Threading a context through `Run` is the real answer if you would rather do it once.

**3. `doctor` and `proxmox status` now disagree about the same endpoint**, which is what the comment above `openProxmoxEndpoint` says must not happen.

`checkProxmoxEndpoint` re-implements `DefaultView` — the same three calls in the same order with the same empty check — but `DefaultView` records the empty case as a warning (`no resources visible; check Proxmox token permissions`, client.go:364) and in `Failed`, which is what `proxmox status`, the web dashboard and the `proxmox_status` MCP tool all render. Against a token whose ACL grants nothing, `doctor` says reachable while those three report the resources collector as failed. Calling `DefaultView` removes the divergence and the duplicated call sequence in one move.

**Smaller things**

- `firstErr` is the first error by collector order rather than the first that classifies, and `httpFailureClass` returns nothing outside 401/403. A `pveproxy` restart that 500s on `/version` while the other two 403 produces the action "Run homebutler proxmox status for the full error" — doctor sending the operator somewhere else, with two authorization failures sitting in the detail it just printed.
- The `len(failed) > 0` branch drops `empties`. A token that lacks `Sys.Audit` and has no resource ACL reports only the cluster-status failure; the operator fixes that, re-runs, and meets the second half for the first time.
- `command` concatenates `endpoint.Name` unquoted, and `loadConfig` does not validate, so doctor runs on configs `config.Validate` would reject. An unnamed endpoint prints `--endpoint` with nothing after it; a name with a space prints something cobra rejects as a stray argument.
- `TestCheckProxmox_TokenFileUnreadableIsFail` asserts `fixture-token` is absent from the detail, but that endpoint carries no token value at all — the string exists only in `openFnFor`, which the test never calls. An inline token makes the assertion guard the property it is named for.
- No CHANGELOG entry, and `doctor`'s scope is enumerated in four places that are now stale: `cmd/doctor.go`'s `Long`, README:97 and :137, and `docs/mcp-server.md:81`. Gaining outbound network calls — and a `fail` for a host that is merely rebooting — is a `⚠️ Behavior changes` line, because `--strict` cron jobs start exiting non-zero for reasons that have nothing to do with the local machine.

CI is green. I read the branch rather than running it: the classification above is from this code and from `tlsConfig` as merged in #120, not from a live endpoint.

Of the three you opened, this was the right one to open first — it is the one that turns #111 from an internal API into something a person sees.

gsaraiva2109:
Fixed all three, plus the smaller things.

1. `tlsConfig`'s CA-file and fingerprint failures now classify `FailureTLS` instead of `FailureAuthentication`. The `FailureTLS` action text — "check the fingerprint or CA file against the endpoint's actual certificate" — already said the right thing, it just never got used for these two error paths.

2. `checkProxmoxEndpoint` now calls `client.DefaultView` instead of reimplementing the three requests. That closes the divergence with `proxmox status` directly: an empty resources response is a warn there, and now here too. It also picks up `DefaultView`'s new early-exit, which covers the network issue — `DefaultView` now gives up after the first `FailureTransport` instead of paying the dial timeout three times, so an unreachable host costs one collector's worth of delay, not three. I didn't thread a context through `Run`; the early-exit gets the practical win without touching every caller (`cmd`, `helpers`, `mcp/server`). Say the word if you'd rather have cancellation end to end.

3. Picked up while in there: `DefaultView`'s first error now prefers a classified error over an earlier unclassified one, so a 500 on `/version` next to a 403 on `/cluster/status` points at the authorization action instead of "run proxmox status for the full error." The suggested `--endpoint` command is quoted. `TestCheckProxmox_TokenFileUnreadableIsFail` now writes a real (permission-denied) secret to a temp file instead of asserting a string that only ever lived in `openFnFor`.

Added the CHANGELOG entry and updated the four stale `doctor` scope descriptions (`cmd/doctor.go`, README x2, `docs/mcp-server.md`).

`gofmt -w .`, `go vet ./...`, `golangci-lint run`, `go test ./...`, `go build ./...` all pass.

Higangssh (approved):
All three, and the two I said were optional as well.

**The class fix is a better answer than either of the ones I offered.** I suggested a new class for configuration material or leaving construction errors unclassified. You noticed that `FailureTLS`'s action text already says "check the fingerprint or CA file against the endpoint's actual certificate" — which is exactly right for a missing CA file or a malformed fingerprint — so the class was the thing that was wrong, not the vocabulary. Nothing new to name and one line moved.

That also unwinds a change I asked for in #120 and then found the cost of here. Sorry for the round trip.

Calling `DefaultView` closes the divergence at the source rather than papering over it, and the early exit is a better shape than the short-circuit I described: `skip` marks the untried collectors as failed, so `proxmox status`, the dashboard and `doctor` all report them as unavailable instead of silently empty, and an unreachable host costs one dial timeout rather than three. Not threading a context through `Run` is the right call for now — the practical win without touching every caller.

`fail` preferring the first *classified* error is the ordering fix, and using `view.Warnings` for the detail means the empty-collector case is carried rather than dropped. `%q` on the endpoint name. And the token-leak test now writes a real secret and asserts it is absent, where before it was asserting against a string that could not have been there.

The `⚠️ Behavior changes` line is the one I care most about in the entry, and it says the thing that costs someone real time: `--strict` cron jobs now exit non-zero for a Proxmox host that is merely rebooting.

`gofmt`, `go vet` and the `doctor`, `proxmox`, `cmd` and `mcp` tests pass locally on `f8288c3`; CI is green.

Thanks — three PRs in and the pattern is that the second version is better than what the review asked for.

Higangssh:
Pushed the CHANGELOG merge onto your branch — main gained an `## [Unreleased]` section from #132 while this was open and the two headings collided. Nothing of yours changed; the entries now sit under one heading. Merging once CI settles.

Issue #97, opened 2026-08-29
watch and alerts are one feature split across two processes


Higangssh:
## Description

homebutler monitors a server with two long-running loops that do not know about each other. Answering "tell me when something goes wrong" currently takes two processes, two config sections, two history stores and two notification paths — and once #80 lands, two service units.

The codebase already says this is a problem in two places.

**`alerts` sends its own users away.** Its `Long` text (`cmd/alerts.go`) reads:

> This is an advanced threshold-based workflow. For most users, start with `homebutler watch` for restart detection, crash analysis, flapping detection, and incident history.

A command whose help tells most people to use a different command is not an advanced mode; it is a split that documentation is covering for.

**`watch` reaches into `alerts`' config to work.** `cmd/watch.go:386`:

```go
if providers.IsEmpty() {
    if alertsCfg, err := loadAlertsConfig(""); err == nil && alertsCfg != nil {
        providers = alerts.ResolveNotifyConfig(alertsCfg)
    }
}
```

`watch start` falls back to reading the alerts config file to find out where to send notifications. That is not two features sharing a library; it is one feature that was cut in half.

## What is actually different, and what is not

The detection models are genuinely different and should stay that way. `watch` is event-driven — it reads `docker events` and compares state across polls to find a *change*. `alerts` polls and reports a *threshold being crossed*. Collapsing those into one configuration surface would make both worse.

What has no reason to be duplicated is everything underneath:

| Duplicated today | Should be |
| --- | --- |
| Two long-running processes | One |
| Two notification paths | One |
| Two history stores | One |
| Two service units, after #80 | One |

## Consequence worth naming

Remediation lives only on the `alerts` side. `executeRestart` runs from the rules engine, which means the loop that detects a container crash still cannot act on one, and the loop that can act cannot see crashes. #49 fixed which target kinds `restart` understands; it did not connect the two halves, because they are in different processes.

Unifying the runtime resolves that without new code: the detector and the hands end up in the same loop.

## Scope

`watch start` becomes the monitoring process. It runs, in one loop with one notifier:

- the existing restart monitors for docker, systemd and pm2
- resource threshold checks, which already have defaults (CPU 90, memory 85, disk 90) so they need no configuration to be useful
- the rules engine, when rules are configured, so remediation reaches restart incidents

[...]

PR #124, opened 2026-09-01
watch: add read-only Proxmox endpoint and guest-state incidents

gsaraiva2109:
## Summary

`watch start` now polls every configured Proxmox endpoint alongside its docker/systemd/pm2 targets, using the same unified monitoring loop from #97 rather than a second scheduler. It reports endpoint reachability and, for guests explicitly named under a new `guests:` key on each `proxmox:` entry, whether an expected-running guest stopped — one incident on the transition, one recovery incident when it clears.

TLS, authentication, authorization, and transport failures keep the classes #111 already introduced on `proxmox.Client`. A response that authenticates but returns no nodes, guests, or storage at all is folded into the same "ACL-filtered" incident state as an outright 403, with the specific failure kept on the incident as `proxmox_class` so the two stay distinguishable in output.

A guest not listed under `guests:` is observational only — never alerted on, deliberately stopped or not. A failed collector on one endpoint doesn't touch guest state already read from another poll or another endpoint.

## Implementation notes

- `internal/watch/proxmox_monitor.go` is a new `ProxmoxMonitor`, structured like `SystemdMonitor`/`PM2Monitor` (seed pass without emitting, then poll on the shared interval) but keyed by endpoint+guest rather than `Target`, since Proxmox targets come from `config.Proxmox`, not the watch target list.
- `ExpectedGuest` lives in `internal/proxmox`, not `internal/config`, because `internal/config` already imports `internal/watch` (for `WatchRuntimeConfig`) — putting it in `internal/config` would have made `internal/watch` importing `internal/config` a cycle.
- `Incident` gained `Source`, `ProxmoxState`, `ProxmoxClass`, and `Recovered` fields rather than a new incident type, so history/show/notify keep working unchanged. `cmd/watch.go`'s print loop branches on `Source == "proxmox"` to skip crash analysis and flapping, which don't mean anything for an API poll.
- All requests are GETs (`client.Resources`); no guest actions, retries, or task polling.

## Test plan

- [x] `gofmt -w .`
- [x] `go vet ./...`
- [x] `golangci-lint run` (no issues)
- [x] `go test ./...` (1217 passed)
- [x] `go build ./...`
- New fixtures in `internal/watch/proxmox_monitor_test.go` cover: guest down/recovered, unconfigured guest never alerting, endpoint unavailable/recovered, ACL-filtered via empty result and via 403, no false alert on an already-stopped guest at startup, and partial-failure preserving guest state across a failed poll.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,13 @@ thresholds, and runs any remediation rules you have configured — one process,
 one set of notification providers. `alerts --watch` still exists and does the
 threshold half on its own.
 
+Every endpoint under `proxmox:` in your config is polled too: unreachable or
+ACL-filtered endpoints and any guest listed under that endpoint's `guests:`
+report one incident when the problem starts and one recovery incident when it
+clears. A guest not listed there is observational only — `watch start` never
+alerts on it, deliberately stopped or not. See
+[Proxmox setup →](docs/proxmox.md#watch-integration) for the `guests:` field.
+
 When a crash is detected, you'll see:
 
 ```
--- a/cmd/proxmox.go
+++ b/cmd/proxmox.go
@@ -25,26 +25,32 @@ func buildProxmoxTargets() []watch.ProxmoxTarget {
 	}
 	targets := make([]watch.ProxmoxTarget, 0, len(cfg.Proxmox))
 	for _, endpoint := range cfg.Proxmox {
+		guests := make([]proxmox.ExpectedGuest, 0, len(endpoint.Guests))
 		for _, guest := range endpoint.Guests {
 			if err := guest.Validate(); err != nil {
 				fmt.Fprintf(os.Stderr, "warning: proxmox endpoint %q: %v\n", endpoint.Name, err)
+				continue
 			}
+			guests = append(guests, guest)
 		}
 
-		token, err := endpoint.TokenValue()
+		newClient := func() (*proxmox.Client, error) {
+			token, err := endpoint.TokenValue()
+			if err != nil {
+				return nil, err
+			}
+			return proxmox.New(proxmox.Options{
+				Host: endpoint.Host, Port: endpoint.APIPort(), TokenID: endpoint.TokenID, Token: token,
+				Fingerprint: endpoint.Fingerprint, CAFile: endpoint.CAFile, Insecure: endpoint.Insecure, Timeout: endpoint.TimeoutDuration(),
+			})
+		}
+		client, err := newClient()
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "warning: proxmox endpoint %q: %v\n", endpoint.Name, err)
+			targets = append(targets, watch.ProxmoxTarget{Endpoint: endpoint.Name, Err: err, NewClient: newClient, Guests: guests})
 			continue
 		}
-		client, err := proxmox.New(proxmox.Options{
-			Host: endpoint.Host, Port: endpoint.APIPort(), TokenID: endpoint.TokenID, Token: token,
-			Fingerprint: endpoint.Fingerprint, CAFile: endpoint.CAFile, Insecure: endpoint.Insecure, Timeout: endpoint.TimeoutDuration(),
-		})
-		if err != nil {
-			fmt.Fprintf(os.Stderr, "warning: proxmox endpoint %q: %v\n", endpoint.Name, err)
-			continue
-		}
-		targets = append(targets, watch.ProxmoxTarget{Endpoint: endpoint.Name, Client: client, Guests: endpoint.Guests})
+		targets = append(targets, watch.ProxmoxTarget{Endpoint: endpoint.Name, Client: client, NewClient: newClient, Guests: guests})
 	}
 	return targets
 }
--- a/cmd/proxmox.go
+++ b/cmd/proxmox.go
@@ -4,12 +4,50 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/Higangssh/homebutler/internal/config"
 	"github.com/Higangssh/homebutler/internal/proxmox"
+	"github.com/Higangssh/homebutler/internal/watch"
 	"github.com/spf13/cobra"
 )
+
+// buildProxmoxTargets turns the global config's Proxmox endpoints into
+// ProxmoxMonitor targets for `watch start`. An endpoint whose client cannot
+// be built (bad token file, invalid TLS setting) is skipped with a warning
+// rather than aborting the whole monitoring run over one misconfigured
+// endpoint — the same tolerance `watch start` already gives an empty target
+// list.
+func buildProxmoxTargets() []watch.ProxmoxTarget {
+	if cfg == nil {
+		return nil
+	}
+	targets := make([]watch.ProxmoxTarget, 0, len(cfg.Proxmox))
+	for _, endpoint := range cfg.Proxmox {
+		for _, guest := range endpoint.Guests {
+			if err := guest.Validate(); err != nil {
+				fmt.Fprintf(os.Stderr, "warning: proxmox endpoint %q: %v\n", endpoint.Name, err)
+			}
+		}
+
+		token, err := endpoint.TokenValue()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "warning: proxmox endpoint %q: %v\n", endpoint.Name, err)
+			continue
+		}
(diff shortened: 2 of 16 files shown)


Higangssh (changes requested):
The structural decision is right and it is the one I would have argued for: polling Proxmox inside the unified loop from #97 rather than standing up a second scheduler, and emitting on transition rather than on every poll. Leaving guest state exactly as it was when a collector fails, instead of guessing, is the correct reading of #104 — and it is what makes the first problem below visible rather than hidden.

Three things before this merges.

**1. The empty-result path does the opposite of the 403 path for guests.**

The description says an authenticated response with no nodes, guests or storage is folded into the same ACL-filtered state as an outright 403. For the endpoint incident it is. For guests it is not, and the difference is `err`:

```go
case len(resources.Nodes) == 0 && ...:
        curr = proxmoxEndpointState{state: ProxmoxStateACLFiltered, class: ProxmoxClassEmptyResult}
...
if err != nil {
        continue   // 403 stops here, guest state preserved
}
```

The empty-result branch is reached with `err == nil`, so the guard does not fire, `running` is built from an empty list, and every configured guest reads as stopped. An ACL that filters everything produces one endpoint incident plus a false `guest_down` for each configured guest — and a false recovery for each when the ACL is fixed. A 403 with the same meaning produces one incident.

**2. No Proxmox incident reaches a provider in the default configuration.**

`cmd/watch.go` calls `notifier.NotifyIncident(inc, watch.FlappingResult{}, nil, ...)`. A zero `FlappingResult` reduces the gate in `NotifyIncident` to `wn.Settings.OnIncident`, and `NotifySettings.Normalize` maps the default `notify_on: "flapping"` to `OnIncident = false`. So on a default watch config the alerting half of this feature is off: the incident is written to disk and printed to stdout, and no provider is ever told.

Skipping flapping analysis for a state transition is the right call — the reason in the comment is sound. But it has to reach the dispatcher some other way, because "flapping" is the default and a Proxmox endpoint going down is exactly the event someone configured a provider for.

**3. A recovery is indistinguishable from a failure once it leaves the monitor.**

The event is built with `Fingerprint: kind + ":" + inc.Container`, so a recovery carries the same fingerprint as the failure it clears and the dispatcher's cooldown drops it for any outage shorter than that window — the case where a recovery matters most. `Status` is hardcoded `"restart"`, and neither `Recovered` nor `PostLogs` reaches the event, so a provider that does get one has nothing to distinguish "pve is down" from "pve is back".

`watch history` and `watch show` have the same gap: neither surfaces `Recovered` or `ProxmoxState`, and the message lives in `PostLogs`, which `History` blanks without `--logs`.

**Also worth fixing here**

- `isRunning` is a plain map lookup, so a guest that is absent from the response — migrated to another node, or dropped by a per-VM ACL change — reads as stopped, alerts, and never recovers. Absent and stopped are different answers and only one of them is about the guest.
- `guest.Validate()`'s error is printed to stderr and the entry stays in the list. Its key can never match a resource key, so it can never fire an incident: the operator is told once, at startup, that a VM they asked to watch is malformed, and then the process runs for weeks silently not watching it. Skip the entry, or refuse to start.
- Nothing damps a flapping endpoint. One hovering at its timeout writes an incident file on every poll — around 120 an hour at the default interval — into the shared incidents directory, where they prune real restart history away under the 200-incident cap. Flapping analysis is deliberately skipped for these, so there is nothing else to absorb it.
- `internal/config/validate.go`'s `checkProxmox` never walks the new `guests:` list, so `config validate` calls a malformed guest entry valid and the only signal is the startup warning above.
- An endpoint whose credential cannot be resolved at startup is dropped for the lifetime of the process with one line on stderr. A revoked token then reads as "not monitored" rather than as the incident this feature exists to report.
- `docs/proxmox.md`, the second bullet of the new Watch integration section, has no main clause — "recorded as ACL-filtered, since the same symptom, an empty resource list, that `proxmox status` already treats as a permissions problem rather than a genuinely empty cluster".

CI is green. I confirmed the first two by reading the branch against `internal/watch/notify.go` and `watchconfig.go` on `main`, not by running a cluster.

One process note, since three PRs arrived together: #124 and #125 both add keys to the `proxmox:` entry and both touch `internal/config/config.go`, `cmd/proxmox.go` and `docs/proxmox.md`. Whichever goes second will need a rebase, so I intend to take them in the order #123, #124, #125 — this one before #125 because it is the larger rebase to inherit. Nothing for you to do beyond knowing why #125 may sit for a bit.

gsaraiva2109:
Addressed in c73f276 and 5a8089a after rebasing onto current main.

- Empty ACL-filtered results now preserve configured guest state, matching the 403 path.
- Proxmox transitions notify providers under the default flapping configuration, with distinct recovery status, fingerprint, and message. History and show expose the Proxmox state and recovery information.
- Missing guests preserve their last observed state. Invalid guest entries are skipped at startup and rejected by config validate.
- Endpoint transitions use the configured flapping thresholds and coalesce repeated events. Credential construction failures remain monitored as incidents, and authentication failures retry credential resolution on the next poll.
- Fixed the documentation sentence and resolved the rebase conflicts with the action-credential changes.

Focused tests, the local Go suite, and all CI checks pass.

Higangssh (changes requested):
Eight of the nine, and the three that mattered are right.

The guest loop is guarded on the ACL-filtered state as well as on `err`, so an empty result now behaves the way a 403 does instead of reporting every configured guest down. A guest missing from the response is `found`-checked and keeps its last state, which was the other half of the same mistake. Endpoint transitions go through the flapping thresholds and coalesce, a recovery gets its own kind and fingerprint so the cooldown cannot eat it, `history` and `show` print `ProxmoxState` and `Recovered`, `config validate` walks `guests:`, a malformed entry is skipped rather than warned about and then monitored, and the docs sentence has a main clause.

**One thing before merging, and it is in the fix rather than in what it fixed.**

Making Proxmox notify under the default configuration is right — a `flapping` default should not mean an endpoint going down is silent. But the gate it went through was the one that also implements `notify_on: "off"`:

```go
shouldNotify := inc.Source == "proxmox" || (flap.IsFlapping && wn.Settings.OnFlapping) || ...
```

`Normalize` maps `off` to `OnIncident = false, OnFlapping = false` and deliberately leaves `Enabled` alone, so a Proxmox incident now short-circuits past a setting whose entire meaning is "do not send me anything". I ran it:

```
notify_on=off → Enabled=true OnIncident=false OnFlapping=false
notify_on: off still sent 1 notification(s) for a proxmox incident
```

Gating on appetite rather than on shape keeps both properties:

```go
(inc.Source == "proxmox" && (wn.Settings.OnIncident || wn.Settings.OnFlapping))
```

`flapping` and `incident` and `all` all notify, because each says the operator wants to hear about something; `off` stays off. A test asserting silence under `off` would be worth having next to the one asserting delivery under the default — the second is what this PR fixed, and the first is what it broke.

`gofmt`, `go vet` and the `watch`, `config`, `cmd` and `proxmox` tests pass locally on `5a8089a`; CI is green.

Thanks for taking the nine and for rebasing through the action-credential changes that landed underneath you. This was the heaviest of the four and the only one where the first version had a problem in every layer it touched.

gsaraiva2109:
Fixed in 8d0f50c.

Proxmox notifications now honor `notify_on: off` while still delivering under the default `flapping` setting. Added coverage for the off case.

Verified with `go test ./internal/watch ./internal/config ./cmd` and `go vet ./internal/watch`.

Higangssh (approved):
That is it. Gating on appetite rather than on shape, and `TestNotifyIncident_ProxmoxRespectsNotifyOff` pins the half that was broken by the fix to the other half.

I ran every value of `notify_on` against a Proxmox incident rather than only the two we had been talking about:

```
notify_on=off      OnIncident=false OnFlapping=false → delivered=false
notify_on=flapping OnIncident=false OnFlapping=true  → delivered=true
notify_on=incident OnIncident=true  OnFlapping=false → delivered=true
notify_on=all      OnIncident=true  OnFlapping=true  → delivered=true
```

and checked that a non-flapping docker restart under `notify_on: flapping` is still silent, so the change did not widen anything but the Proxmox path.

`gofmt`, `go vet` and the `watch`, `config`, `cmd`, `proxmox` and `report` tests pass locally on `8d0f50c`; CI is green.

Thanks for four PRs and for this one in particular — nine findings, a rebase through the action-credential work landing underneath it, and then a tenth that only existed because of the ninth. The Proxmox set is finished apart from #106.

Issue #107, opened 2026-08-30
security: optionally separate Proxmox read and action credentials


gsaraiva2109:
## Problem

A single Proxmox endpoint credential is simple, but it couples dashboard, watcher, `doctor`, read-only MCP tools, and guest power actions. A credential used continuously for observation should not need power-management permission.

## Proposal

Support an optional second, narrowly scoped action credential per Proxmox endpoint. The standard endpoint credential remains read-only. Dashboard, watcher, `doctor`, status commands, and read-only MCP tools always use only that credential. Guest start, reboot, and graceful shutdown use the action credential only when it is explicitly configured.

If no action credential is configured, guest actions are unavailable even if the read credential happens to have power permission. There is no silent fallback.

## Discussion required before implementation

This is intentionally a design issue. Do not start implementation until these decisions are settled:

1. Configuration shape: separate endpoint fields, a nested `actions:` block, or separate named action endpoint entries.
2. Backward compatibility: existing single-token action setups currently work. Should they continue temporarily with an explicit warning, require migration, or be gated behind an opt-in compatibility setting?
3. Token storage: whether action tokens reuse `token_file` conventions with a dedicated owner-only file, and how errors remain secret-safe.
4. Permission model: documented minimum read ACLs plus per-guest `VM.PowerMgmt` action ACLs, including Proxmox privilege-separated token intersection.
5. UX: how CLI, MCP, and dashboard explain that actions are unavailable because no action credential is configured.
6. Rotation and revocation: what should fail independently, and what `doctor` should validate without printing a secret.

## Boundaries

- Opt-in hardening, not a mandatory setup requirement.
- No global `Administrator`, wildcard action permission, hidden fallback credential, or shared token value in output.
- No automatic guest action, retries of state-changing calls, SSH fan-out, or dashboard action controls.
- Keep existing explicit endpoint, node, guest type, VMID, action, and confirmation requirements.

## Done when

After the discussion resolves the model, a follow-up implementation can prove:

- observation paths cannot use the action credential
- action paths fail safely when no action credential is configured
- read and action token failures are reported separately without credential leakage
- the documented Proxmox ACLs are least privilege and guest-scoped

Related: #32 for the Proxmox integration, #90 for guest actions, #105 for diagnostics, and #106 for dashboard status.

Higangssh:
[...]

PR #125, opened 2026-09-01
feat: require a separate credential for proxmox guest actions

gsaraiva2109:
A single Proxmox token has done double duty. The same credential that drives read only status calls also carries the power to start, reboot, and shut down a guest, even though status polling never needed that permission in the first place. This closes issue #107, where that coupling was flagged as a design problem worth fixing before it caused an actual incident.

Guest start, reboot, and shutdown now resolve a second, optional credential (action_token_id plus action_token or action_token_file) instead of the endpoint's read token. The dashboard, watcher, and read only MCP tools are untouched, since they never needed anything more than the read credential.

There is no fallback and no compatibility flag. Anyone already running guest actions today needs to add the new fields to their config, or those commands start failing with a named error the moment this ships. That tradeoff was made deliberately: a warning that keeps the old behavior working would reintroduce the exact silent fallback the issue exists to remove. docs/proxmox.md walks through creating the second token with a least privilege ACL, including a note on the token and user permission intersection, since reusing the read user for the action token can quietly widen its access.

On the implementation side, config resolution for both credentials now goes through one method, ProxmoxConfig.ResolveCredential, so the CLI and MCP paths share the same logic and the same error text instead of each hand rolling the branch.

Verification:
- gofmt -w .
- go vet ./...
- golangci-lint run
- go test ./...
- go build ./...

Full design discussion, including the answers to the issue's six open questions, is in the comment thread on #107.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+**Guest power actions now need their own Proxmox credential, separate from the one that reads status.** The same token drove both, so anything with read access to the config also held `VM.PowerMgmt` on every guest, even though status polling never needed it.
+
+### ✨ Features
+
+- require a separate credential for Proxmox guest actions (#107). `action_token_id` plus `action_token` or `action_token_file` configures a second, optional token used only for start, reboot, and shutdown; reads, the dashboard, and the read-only MCP tools keep using the existing token. `config validate` warns when the action credential is the same token as the read one, since that defeats the separation. See [Proxmox setup →](docs/proxmox.md) for creating the second token with a least-privilege ACL
+
+### ⚠️ Behavior changes
+
+- **Guest start, reboot, and shutdown fail with `no action credential configured for Proxmox endpoint "..."` until `action_token_id` and `action_token` (or `action_token_file`) are added to the endpoint.** There is no fallback to the read token and no compatibility flag — reusing it defeated the reason this credential is separate in the first place
+
 ## [0.25.0](https://github.com/Higangssh/homebutler/compare/v0.24.0...v0.25.0) - 2026-09-02
 
 **Every permission failure homebutler can hit printed the operator's next command last.** `backup`, `restore`, `install` and `upgrade` all led with the raw syscall error and put the one line that mattered underneath it — and the detail could not simply be dropped, because there was no way to ask for it back. This release inverts the order and adds the flag that makes dropping it safe.
--- a/README.md
+++ b/README.md
@@ -459,10 +459,14 @@ CA file, and only then an explicit `insecure` fallback.
 
 Reads are plain. Power actions are not: every one of them takes an explicit
 endpoint, node, guest type and VMID, and refuses to run without `--confirm`,
-which is checked before the token is even read. `shutdown` asks the guest to shut
-down cleanly — it is not Proxmox's hard `stop`, which cuts power and can leave a
-filesystem behind it. A successful action reports the task it submitted, not that
-the guest finished; `proxmox task` answers that separately.
+which is checked before any credential is read. They also need their own
+`action_token_id` (plus `action_token` or `action_token_file`) configured on
+the endpoint — the read token alone will not start, reboot, or shut down a
+guest; see [Proxmox setup →](docs/proxmox.md) for creating that second token.
+`shutdown` asks the guest to shut down cleanly — it is not Proxmox's hard
+`stop`, which cuts power and can leave a filesystem behind it. A successful
+action reports the task it submitted, not that the guest finished; `proxmox
+task` answers that separately.
 
 `proxmox script` prints the install command for a Community Script pinned to one
 commit, along with a warning that the script is not reviewed by homebutler and
--- a/cmd/proxmox.go
+++ b/cmd/proxmox.go
@@ -45,6 +45,18 @@ func newProxmoxStatusCmd() *cobra.Command {
 }
 
 func openProxmoxClient(endpointName string) (*config.ProxmoxConfig, *proxmox.Client, error) {
+	return openProxmoxClientCredential(endpointName, false)
+}
+
+// openProxmoxActionClient is like openProxmoxClient but authenticates with the
+// endpoint's action credential instead of its read credential. Guest power
+// actions must never fall back to the read token: if no action credential is
+// configured, this fails before any network call.
+func openProxmoxActionClient(endpointName string) (*config.ProxmoxConfig, *proxmox.Client, error) {
+	return openProxmoxClientCredential(endpointName, true)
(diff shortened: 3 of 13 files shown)


Higangssh (changes requested):
The security property this PR claims actually holds, and I traced it rather than taking it. `isAction` is derived only from the tool-name switch and the guest-action cobra command, so no read path can pick up the action token and no action path can fall back to the read one. `--confirm` is still checked before any credential is resolved on both the CLI and MCP sides, which keeps #90's ordering. The `secret:"true"` tag means the reflective `hasSecrets` walk covers `ActionToken` without anything being wired by hand, and the masking in `Client` covers it too. Routing both credentials through `ResolveCredential` is the right shape.

**On the breaking change: that call is mine, and I am making it the same way you did.** No fallback is correct. A warning that kept the old behaviour working would leave exactly the silent coupling #107 exists to remove, and pre-1.0 is when this costs least. It does have to be visible, which is the first of the two things below.

**One change before merging.** `config validate` accepts an action credential that is byte-identical to the read one.

The new block checks presence, exclusivity and file existence, and nothing compares the two credentials. Put yourself at the other end of the new error message: `homebutler proxmox guest shutdown` starts failing with `no action credential configured`, and the fastest thing that makes it stop is copying `token_id` and `token_file` into `action_token_id` and `action_token_file`. Validation then reports the config as valid, the same token carries `VM.PowerMgmt` again, and the operator believes they have the separation this release told them to set up.

That is the one misconfiguration that defeats the whole feature, and it is the one the error message steers people toward. A `SeverityWarning` when `action_token_id == token_id`, or the token values or file paths match, closes it.

**And the remediation hint contradicts the documentation in the same diff.**

`validate.go` suggests `monitoring@pve!action` — an action token owned by the read user. `docs/proxmox.md`, added in this PR, warns against precisely that:

> Reusing the read token's user (which already holds `PVEAuditor`) as the action token's owning user … can silently widen the action credential's effective access beyond `VM.PowerMgmt`

and its own worked example uses a dedicated `power@pve`. The hint is the copy most people will act on, because it appears at the moment they are fixing the error. It should name a dedicated user.

**Smaller things**

- No CHANGELOG entry. This is the case the `⚠️ Behavior changes` section exists for: an operator who upgrades and meets `no action credential configured for Proxmox endpoint "pve"` currently has nothing to read that explains why. #121 added one for a far smaller change.
- README:460 still says power actions need `--confirm` and "the token", and the quick-tour `proxmox guest shutdown` example above it now fails for every reader until they configure a second credential.
- `TokenValue` and `ActionTokenValue` classify their failures as `FailureAuthentication`, but the new "no action credential configured" error is a bare `fmt.Errorf`, so `Classify` returns nothing for it. #111 landed one release ago so callers could branch without matching text, and this will be the most common action failure there is.
- `internal/server/server.go:497` still calls `endpoint.TokenValue()` and wraps by hand. It is correct today and deliberately generic in what it returns, but it is the one call site left outside the single method this PR introduced, so it is the one that drifts. `ResolveCredential(false)` reads the same and keeps the log line.
- `action_token_file` is checked for existence but not permissions, mirroring `token_file`. That was fine when the file behind it was audit-only; it now grants `VM.PowerMgmt`, and `config.Load` already refuses `0o077` bits on a config holding plaintext secrets. A warning on a group- or world-readable action token file closes the same hole for the credential that can power-cycle a guest.

CI is green, and I read the branch rather than running it against a cluster — the credential routing above is traced through the call sites, not observed.

Thanks for writing the design discussion into #107 rather than into the PR body. It is the second time this week the reasoning outlived the diff, and it is why this one was quick to check.

On sequencing: this one shares `internal/config/config.go`, `cmd/proxmox.go` and `docs/proxmox.md` with #124, and both add keys to the same `proxmox:` entry. I am taking #123 and #124 first, so expect a rebase here rather than a conflict.

gsaraiva2109:
Addressed everything in the review, in `a600adf`:

- **Identical-credential gap**: `checkProxmox` now warns (`SeverityWarning`) when `action_token_id == token_id`, or the action token's inline value / file path matches the read credential's. Copying the read credential into the action fields no longer validates clean.
- **Remediation hint**: `action_token_id` hint now points at a dedicated user (`power@pve!action`), matching `docs/proxmox.md` instead of contradicting it.
- **No CHANGELOG entry**: added an `## [Unreleased]` section noting the new required action credential and the `no action credential configured` behavior change.
- **README:460**: updated to say power actions need their own `action_token_id`/`action_token`(`_file`) — the read token alone won't run them — and pointed at `docs/proxmox.md` for setup.
- **Unclassified error**: "no action credential configured" is now wrapped with `proxmox.WithFailureClass(proxmox.FailureAuthentication, ...)`, so `Classify` returns something callers can branch on, same as `TokenValue`/`ActionTokenValue`.
- **server.go:497 drift**: `handleProxmoxStatus` now goes through `ResolveCredential(false)` instead of calling `TokenValue()` by hand.
- **action_token_file permissions**: added the same `0o077` group/other-readable warning that `token_file` already gets, scoped to the action file.

Added tests for all of the above (four new cases in `validate_test.go`, one assertion added to the existing `ResolveCredential` test in `config_test.go`). Full suite (`gofmt`, `go vet`, `golangci-lint`, `go test`, `go build`) is clean.

Ready for the rebase once #123/#124 land.

Higangssh (approved):
Everything, and the two I marked optional.

**The identical-credential check is the one that mattered** and it covers both doors: `action_token_id == token_id` warns, and so does an action token whose inline value or file path is the read credential's. Copying the read credential into the action fields to make the new error go away — the fastest thing anyone will try — no longer validates clean. `sameProxmoxCredential` being explicit that it compares paths rather than contents is the right scope: two paths can hold the same secret, and config alone cannot know.

The hint names `power@pve!action` now, so it stops contradicting the documentation added in the same diff.

`server.go` going through `ResolveCredential` closes the last hand-rolled call site. I checked the thing that worried me — `log.Print(err)` losing the endpoint name — and it does not: `ResolveCredential` names the endpoint in every one of its errors, so the log line still says which one failed.

The permission warning on `action_token_file` was a take-it-or-leave-it and I am glad you took it. That file now grants `VM.PowerMgmt`, and `config.Load` already refuses `0o077` on a config holding plaintext secrets; this closes the same hole for the credential that can power-cycle a guest.

`gofmt`, `go vet` and the `config`, `mcp`, `cmd` and `proxmox` tests pass locally on `a600adf`; CI is green.

The breaking change stands as we agreed: no fallback, and the changelog says so plainly. Thanks for writing the design discussion where it belonged and for taking the security half seriously enough to argue for the harder option.

Higangssh:
Pushed the CHANGELOG merge onto your branch — #123 landed while this was open and both had added an `## [Unreleased]` heading. Nothing of yours changed; the entries sit under one heading now. Merging once CI settles.

Issue #59, opened 2026-08-22
processes and network are observed but never tracked


Higangssh:
## Description

`Snapshot` (`internal/report/report.go:21`) holds `System`, `Containers`, `Ports`, and `Warnings`. That is the whole set of things homebutler can compare across runs.

`processes` and `network` are collected and rendered but never reach it. So "what is eating the CPU" can be answered for right now and never compared against last time — the question "what started eating the CPU since yesterday" has no path to an answer, even though the command to observe it has existed for a while.

## Why it matters more than it looks

The README asks "Why did this service restart at 3 AM?" and answers it with `watch`. The neighbouring question — "what changed on this box between then and now" — is `report`'s job, and it is answered only for containers, ports, and disks.

A process that appeared and stayed, or a listening socket that moved, is the kind of thing an operator notices days later and cannot reconstruct. The observation already runs; only the persistence is missing.

#54 adds a `processes` MCP tool, which makes this more visible: an agent will be able to ask what is running now and still have no way to ask what changed.

## Scope

- Add processes to `Snapshot`, with a stable identity to match on across runs. Pid is not it — pids are recycled and change on restart. Likely name plus command, with a decision needed on how much of the command line to keep.
- Decide what "changed" means for a process before implementing the diff. CPU and memory fluctuate constantly; a naive delta would make every report noisy. Appeared and disappeared are the useful signals; resource deltas probably need a threshold or should be left out.
- Same question for `network`, which may not belong in a snapshot at all. Interface state changes are worth recording; per-run throughput numbers are not. Worth settling before writing code.

Snapshot files grow with whatever is added here, and `pruneSnapshots` keeps `Keep` of them. Worth checking the size impact on a busy host before committing to a shape.

## Notes

Depends on the identity-based diff landing first — adding data to the snapshot is only useful once the report compares contents rather than counts.

Higangssh:
Answering the three decisions this issue defers, so it can be picked up.

**Process identity is name plus `argv[0]`.** Not the pid, for the reason already in the issue, and not the full command line — command lines carry secrets in flags, and a snapshot file lives on disk under `~/.homebutler/reports/snapshots/` where nothing has ever had to be handled as a credential. Store a hash of the full command line alongside the identity: that still detects "same binary, different invocation" across runs without persisting the arguments themselves.

[...]

PR #127, opened 2026-09-02
feat: compare containers and listeners by identity, not by count

### LINKED ISSUE #58: report compares counts, so a container swap reads as no change
## Description

The README leads with "Know what changed before you fix it", and `report` is what delivers that. It currently delivers much less than the data it already stores allows.

`Snapshot` (`internal/report/report.go:21`) persists `Containers` and `Ports` in full, as complete lists, to `~/.homebutler/reports/snapshots/`. `buildReport` then compares four things:

- per-mount disk `UsedGB` delta
- `RunningCount`
- `StoppedCount`
- `PublicPortCount`

Three of those four are integers. The lists themselves are written to disk and never read back for comparison.

## What that costs

If `vaultwarden` stops while `gitea` starts, `RunningCount` is unchanged and the report says:

```
No significant changes since last report.
```

The same holds for ports. If `:8080` changes owner, `PublicPortCount` is unchanged and nothing is reported. A container replaced by a different container, a port that moved to a new process — these are exactly the events the command exists to surface, and it stays quiet for all of them.

The previous snapshot file already contains everything needed to catch this. It is read for its counts and then discarded.

## Scope

Diff by identity rather than by cardinality:

- Containers matched by name: appeared, disappeared, changed state, changed image
- Ports matched by port number and protocol: appeared, disappeared, changed owner

Keep the output shape. `Report.NotableChanges` is `[]string`, so a richer diff produces more strings without touching the JSON schema — this does not interact with the 1.0 surface freeze and does not need to wait for it.

Worth deciding as part of this: how much detail belongs in `NotableChanges` before it stops being readable. A server with thirty containers rebooting should not produce sixty lines. Some form of grouping or a cap is probably needed, and that decision is the reason this is not purely mechanical.

## Notes

Highest-priority work on the change-tracking side: it implements the differentiator with zero new observation, using data already on disk.

Related but separate: `processes` and `network` are not in `Snapshot` at all, so they cannot be diffed regardless of how good this gets.

Higangssh:
Closes #58.

The snapshot on disk has held the full container and port lists since the format was written. `buildReport` read four integers off it and threw the lists away, so a container replaced by a different container under the same name left every count identical and the report said `No significant changes since last report`.

Containers are now matched by name, then by ID, image and state; listeners by protocol and port number, then by the process answering. The kind that matters is `replaced` — same name, different container underneath — because it is the one no count can see.

The decisions the issue deferred, and what they came out as:

**What earns a line** is in `docs/report.md`, which is new, along with what does not: uptime and status strings (`Up 4 hours` moves between any two runs), CPU and memory (a single-sample delta is noise, and a real signal needs a threshold and a duration), and restart counters that moved without a state change. `TestVolatileStatusIsSuppressed` pins the first one, because the cheapest way to ruin this feature is to make it noisy.

**Grouping is human-only.** More than five changes of one kind collapse into a line naming three and counting the rest. `--json` carries every change, ungrouped — a person needs the section readable, and quietly handing a shortened list to an agent is how an agent becomes confidently wrong. `NotableChanges` stays `[]string`, so the JSON schema is unchanged and this does not touch the 1.0 freeze.

**A collector that did not answer stops the comparison** rather than reporting everything as gone. `Snapshot.Failed` existed for exactly this and nothing had read it.

Two things came out of running it rather than testing it. Grouping keyed on kind alone produced `gone 11 containers` over a list of ports, since `gone` describes both — `TestCollapsedLinesCountWhatTheyCollapsed` is that bug. And the human report now leads with `Notable Changes`: with the section finally saying something, printing current status above it put the reading every other tool already shows ahead of the one only this one prints.

`assets/report-card.svg` is redrawn to what the command actually prints, and `TestRenderedChangeBlock` pins that block so a format change fails a test rather than quietly dating the image.

Verified by running it: two `report` runs either side of opening a listener, which reported `new :9998/tcp — now Python`, and with Docker stopped, which skipped the container diff and said so. Snapshots written during that were removed afterwards. `gofmt`, `go vet`, `go build` and `go test ./...` are clean locally; `golangci-lint` is not installed here, so CI is the first thing to run it.

#59 is next and depends on this — processes and network are still collected and never persisted.


Higangssh:
Pushed fixes for six things found by re-reading this after opening it — the first is the one that matters.

`indexPorts` dropped the bind address, so `127.0.0.1:8080` and `0.0.0.0:8080` were the same entry. A service moving from loopback to every interface — the change in this section with the largest consequence — reported `No significant changes since last report`, in the same output where Suggested Actions said "New public port(s) detected". It also invented owner changes whenever the platform tool reordered its rows. Verified after the fix by running it:

```
   gone     127.0.0.1:9997/tcp  was Python
   new      :9997/tcp           now Python
```

Also: a replacement that came back down now names the state transition instead of swallowing it; a skipped comparison is a line in the section rather than only a trailing warning, so nothing reading `notable_changes` gets an all-clear the report cannot stand behind; `diffDisks` joins on mount instead of cross-producting; and padding measures display width rather than bytes. `docs/report.md` is linked from the README like every other page, and the stale card filename in the test comment is corrected.

Each has a test named after the symptom.

PR #128, opened 2026-09-02
feat: support modern MCP requests without a handshake

### LINKED ISSUE #86: MCP removed the initialize handshake, and homebutler only speaks the old era
## Description

MCP `2026-07-28` removed the `initialize`/`notifications/initialized` handshake. Every request now carries its protocol version and client capabilities in `_meta`, servers **MUST** implement a `server/discover` RPC returning their supported versions, capabilities and identity, and protocol-level sessions are gone so any request can land on any instance.

homebutler implements the handshake and nothing else (`internal/mcp/server.go:160`). In the spec's own terminology it is a **legacy** server.

## This is not currently broken

The spec publishes a compatibility matrix, and the row that matters says so plainly:

| Client | Server | Outcome |
|---|---|---|
| Legacy | Legacy | Works |
| **Dual-era** | **Legacy** | **Works** — the client falls back to `initialize` |
| Modern-only | Legacy | Fails |

On stdio a dual-era client probes with `server/discover`, gets something that is not a recognised modern error, concludes the server is legacy, and falls back. homebutler's `-32601 method not found` is exactly the signal that probe is looking for, so it works by accident and by design at the same time.

What fails is a client that speaks only the modern era. The revision is weeks old, so that is a question of when, not whether.

## Most of the revision does not apply here

homebutler is stdio-only (`server.go:129`, `:136` — `os.Stdin` and a scanner; there is no HTTP server in the MCP path). That removes a large part of the change from scope: the `Mcp-Session-Id` header, SSE resumability and `Last-Event-ID`, `subscriptions/listen` replacing the GET endpoint, and the required HTTP request headers are all Streamable HTTP concerns.

Also out of scope: Roots, Sampling and Logging are deprecated in this revision, and homebutler implements none of them.

What is actually in scope for a tools-only stdio server:

- `server/discover` returning supported versions, capabilities and identity
- Reading `io.modelcontextprotocol/protocolVersion` from each request's `_meta` and answering `UnsupportedProtocolVersionError` (`-32022`) with a `supported` list when it does not match
- `resultType: "complete"` on results
- `ttlMs` and `cacheScope` on `tools/list`, per the new `CacheableResult` interface — a freshness hint that lets clients stop re-listing
- Deterministic tool ordering from `tools/list`, which the revision recommends for client-side caching and prompt cache hits

## Dual-era rather than a cutover

The spec allows it directly: a server that wants to serve both eras **MAY** implement both behaviours, selecting from how the client opens — a request carrying modern `_meta` is served statelessly, an `initialize` request selects legacy semantics. That keeps every client working today while making the modern path available.

If homebutler ever does go modern-only, the spec asks for one courtesy: name the supported protocol versions in whatever error is returned to an `initialize` request, because legacy clients have no way to fall forward and that message may be the only diagnostic a user sees.

## Why this is a 1.0 question

`CONTRIBUTING.md` says 1.0 freezes the MCP tool surface and the JSON schema. How the server is reached is part of that surface: `server/discover` has to report capabilities, and `resultType` and the cache fields change the shape of every result. Deciding after 1.0 means deciding it as a breaking change.

## Done when

- A modern client can talk to homebutler without a handshake
- A legacy client keeps working unchanged
- The era a client is using is determined by how it opens, not by configuration
- `tools/list` carries a freshness hint and returns a deterministic order

## Notes

#85 covers moving the legacy declaration off `2024-11-05`, which is small and independent. This issue is the larger decision and should not block it.

Spec references: [versioning and compatibility](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning), [key changes](https://modelcontextprotocol.io/specification/2026-07-28/changelog).

lenny-ts:
homebutler now supports MCP 2026-07-28 modern requests while preserving legacy client compatibility.

The implementation adds:

- `server/discover`
- Per-request protocol metadata validation
- `UnsupportedProtocolVersionError` with supported versions
- Stateless modern request handling
- `resultType: "complete"` on modern results
- `ttlMs` and `cacheScope` on `tools/list`
- Deterministic tool ordering
- Legacy `initialize` and `ping` compatibility
- Updated MCP documentation and tests

Verified with:

- `go vet ./...`
- `golangci-lint run`
- `go test ./...`
- `go build ./...`

close #86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,38 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
-**Modern MCP clients can now use homebutler without a lifecycle handshake while legacy clients keep working.**
+**Modern MCP clients can now use homebutler without a lifecycle handshake, while reports now identify what changed between snapshots.**
 
-### Features
+```
+── Notable Changes ───────────────────────────────────────────
+   gone      vaultwarden   was running
+   new       gitea         now running
+   replaced  nginx         7d4a91f0aa11 → 91be0322bb22
+   image     jellyfin      jellyfin:10.9.11 → jellyfin:10.10.0
+   state     7 containers  postgres, redis, grafana, +4
+   port      :8080/tcp     vaultwarden → gitea
+   disk      /             +2.4 GB since last report
+```
+
+### ✨ Features
 
 - add the MCP 2026-07-28 `server/discover` RPC, per-request metadata validation, version mismatch errors, result typing, cache hints, and deterministic tool ordering
+- compare containers and listeners by identity rather than by cardinality (#58). Containers are matched by name and then by container ID, image and state; listeners by bind address, port number and protocol, then by the process answering on them. The address is part of a listener's identity: without it a service moving from `127.0.0.1` to `0.0.0.0` — the change in this section with the largest consequence — reads as no change at all. The case this exists for is `replaced` — a container recreated under the same name, which leaves every count identical and was therefore invisible. `docs/report.md` is new and states what earns a line and what is deliberately suppressed: uptime strings, CPU and memory, and restart counters that moved without a state change, because a report nobody reads is worse than no report
+- collapse more than five changes of one kind into a line that names three and counts the rest. Grouping and truncation happen only in the human renderer — `--json` carries every change, ungrouped, because a person needs the section readable and anything parsing the output needs all of it
+- skip the comparison, and say so *in the section*, when a collector did not answer for either snapshot. Diffing against a snapshot taken while Docker was down would otherwise report every container as gone — and putting the caveat only in a trailing warning would hand anything reading `notable_changes` an all-clear the report cannot stand behind
+- track processes across runs (#59). A process is identified by its executable name and a hash of its full invocation, so a name that appears, disappears, or comes back running something else is reported — `python3 /opt/old.py` becoming `python3 /opt/new.py` is a `replaced` rather than silence. Identical invocations collapse to one identity, because eight workers sharing a command line are one thing running and a pool resizing is a resource signal rather than an arrival
+- ignore processes that have not been running for a minute. This threshold was measured rather than chosen: two runs thirty seconds apart on an idle machine reported `new head` and `new sed`, which was the shell pipeline reading the report. A process held back is reported on the next run, by which time it has earned the word "new"
 
-### Behavior changes
+### 🔐 Security
+
+- keep the executable name and a twelve-character hash of the invocation, never the command line itself. Command lines carry secrets in flags and `~/.homebutler/reports/snapshots/` has never had to be handled as a credential store. `ProcessInfo.Command` is excluded from JSON, so neither the `processes` command nor its MCP tool gains the field
+
+### ⚠️ Behavior changes
 
 - modern requests use stateless per-request metadata; legacy `initialize` requests continue to negotiate legacy protocol versions
+- **`notable_changes` no longer contains the count lines** `Running containers: N → M`, `Stopped containers: N → M` or `Public ports: N → M`. They are replaced by one entry per change, shaped `kind: subject — detail`. Anything matching on the old strings needs updating; the field is still `[]string` and the JSON schema is unchanged
+- **The human report leads with `Notable Changes` rather than `Current Status`.** Current status is context for what moved above it, not the headline. `--json` is unaffected — field order in the document has not changed
+- **Snapshots are larger.** On a desktop macOS with 639 distinct process identities a snapshot is about 50 KB, against 2 KB before; a Linux server running a few services is well under that. At the default `--keep 30` that is roughly 1.5 MB of history in the worst case measured. No cap is imposed — the measurement did not call for one — but it is worth knowing before turning `--keep` up
 
 ## [0.25.0](https://github.com/Higangssh/homebutler/compare/v0.24.0...v0.25.0) - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+**Modern MCP clients can now use homebutler without a lifecycle handshake while legacy clients keep working.**
+
+### Features
+
+- add the MCP 2026-07-28 `server/discover` RPC, per-request metadata validation, version mismatch errors, result typing, cache hints, and deterministic tool ordering
+
+### Behavior changes
+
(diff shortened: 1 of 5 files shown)


Higangssh (changes requested):
This is the shape #86 asked for: the era comes from how the client opens rather than from configuration, and the legacy path is untouched. Several things I would have checked first are right.

The `UnsupportedProtocolVersionError` payload matches the published example field for field — `-32022`, `Unsupported protocol version`, and a `data` carrying both `supported` and `requested`:

```json
{"code":-32022,"message":"Unsupported protocol version",
 "data":{"requested":"1900-01-01","supported":["2026-07-28", ...]}}
```

Answering `-32601` to `ping` under modern metadata is a good catch and not one the issue spelled out. The revision removes `ping`, and #118 added answering it three days ago — keeping it for legacy while refusing it for modern is exactly the dual-era split. `ttlMs` and `cacheScope` go on `tools/list` and not on `tools/call`, which follows the `CacheableResult` list rather than sprinkling them everywhere, and `_meta` carries `serverInfo` on every result as the revision asks.

**One change before merging: a modern client cannot reach the modern path.**

`server/discover` is refused unless the request already carries modern `_meta`:

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"server/discover"}' | homebutler mcp
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"server/discover requires modern request metadata"}}
```

That is the request a dual-era client sends *first*, before it knows which era it is talking to. The stdio binding says to "probe with `server/discover` and fall back on any error that is not a recognized modern error", and the compatibility matrix says a dual-era client against a modern server works because "the stdio probe returns a `DiscoverResult` (or `UnsupportedProtocolVersionError`)". `-32602` is neither, so the client concludes homebutler is legacy and falls back to `initialize` — and everything else in this PR is never used by the one client it was written for.

`server/discover` should answer a `DiscoverResult` whatever the request carries, and return `-32022` only when `_meta` names a version this server does not implement. The probe exists precisely for a client that does not yet know what to put in `_meta`.

**Second: which list governs `_meta` validation.**

`isSupportedProtocolVersion` checks against all five versions, so a request that declares a legacy revision in modern metadata is served as modern:

```
$ ... '"io.modelcontextprotocol/protocolVersion":"2024-11-05"' ... tools/list
{"resultType":"complete","tools":[...],"ttlMs":300000,"cacheScope":"public", ...}
```

`resultType`, `ttlMs` and `cacheScope` do not exist in `2024-11-05`. By the revision's own terminology, modern versions are the ones that carry version and capabilities as per-request metadata — `2026-07-28` and later — so a legacy version arriving in `_meta` is a contradiction and belongs in the `-32022` path.

The split is already half made: `legacyProtocolVersions` exists and `negotiateProtocolVersion` uses it. `_meta` validation wants the other half of that same split. Advertising all five from `server/discover` is right and I would leave it — the spec's own example mixes eras, and a dual-era server saying it can also be reached by `initialize` is true.

**Smaller things**

- No CHANGELOG entry. The expectation went into `CONTRIBUTING.md` this morning in #126, so this is bad timing rather than anything you skipped — but this one needs it more than most, since #86 argues the reachability of the server is part of what 1.0 freezes.
- `ttlMs: 300000` is a bare number in the middle of a struct literal. Five minutes is a reasonable guess; the next reader should be told it is a guess and what would change it.
- `initialize` carrying modern `_meta` is answered `-32601`. That is defensible — the two ways of opening contradict each other — and it is the one case the spec leaves to the implementation, so it is worth a line in `docs/mcp-server.md` saying so rather than leaving it to be discovered.

I checked the four paths above against a build of this branch over stdio rather than reading them, and the spec claims against the published `2026-07-28` versioning page and changelog rather than against #86, since the issue is my summary of the revision and not the revision itself. What I have not done is a second pass over the Go for ordinary correctness — everything above is conformance.

lenny-ts:
Addressed the review feedback:

- `server/discover` now works without `_meta`, enabling the stdio compatibility probe.
- Legacy protocol versions in modern metadata now return `-32022`.
- Documented the `initialize`/modern metadata behavior.
- Added the MCP changes to the `Unreleased` changelog while preserving the complete report entry.
- Added coverage for discovery probing and legacy-version rejection.
- Merged the latest `main` changes into the branch.

All checks pass: `go vet ./...`, `golangci-lint run`, `go test ./...`, and `go build ./...`.

Higangssh (approved):
Both, and I checked them the same way as last time — a build of this branch, over stdio.

The probe answers now:

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"server/discover"}' | homebutler mcp
{"result":{"resultType":"complete","supportedVersions":[...],"capabilities":{"tools":{}},...}}
```

which is what makes everything else in the PR reachable: a dual-era client's first request no longer reads as a legacy error. A legacy version arriving in modern `_meta` gets `-32022`, and `initialize` still negotiates the legacy path untouched, so the dual-era split holds from both sides. `isModernProtocolVersion` being `supported && !legacy` is the right shape — the two lists were always going to drift if the distinction stayed implicit.

`TestDiscoverProbeWorksWithoutMetadata` and `TestLegacyVersionInModernMetadataIsRejected` are named after the failures rather than the functions, which is what makes them survive a refactor.

`gofmt`, `go vet` and the `mcp` and `cmd` tests pass locally on `754ae47`; CI is green.

**One thing I am not holding this for.** The `-32022` payload lists all five versions, including the one it just refused:

```
"error":{"code":-32022,"data":{"requested":"2024-11-05","supported":["2026-07-28","2025-11-25","2025-06-18","2025-03-26","2024-11-05"]}}
```

Literally correct — this server does implement `2024-11-05` — but the spec tells the client to "select a mutually supported version from the `supported` list and retry", and a client that picks the legacy one gets the same error again. The versions usable in `_meta` are the modern ones; `server/discover` advertising all five is right, since a dual-era server genuinely can be reached both ways.

It needs a modern client to send a legacy version in the first place, so it is theoretical rather than something anyone is hitting. I am opening it as a follow-up rather than asking for a fourth round — it is a question about what a dual-era server should advertise in that specific error, and it deserves its own thread before 1.0 freezes the answer.

Thanks for taking the largest of the four and for the spec reading behind it. This is the piece that decides whether homebutler is still reachable in a year.

Higangssh:
Pushed the CHANGELOG merge onto your branch. Two things moved: main's `## [Unreleased]` section is kept whole, and your two MCP entries were sitting inside the 0.26.0 section — that release was cut while this PR was open — so they moved up under Unreleased. Nothing else in the file changed; `git diff origin/main -- CHANGELOG.md` is those two lines and nothing more.

Merging once CI settles.

PR #129, opened 2026-09-02
feat: track processes across runs

### LINKED ISSUE #59: processes and network are observed but never tracked

Higangssh:
Closes #59, and finishes the pair #58 opened.

`processes` ran on every report and was thrown away. The snapshot held containers, ports and disks, so the question in the issue — what started running since yesterday — had no path to an answer even though the observation already happened.

Two of the decisions I left on the issue were wrong about the collectors, and I corrected them there before starting:

**There was no command line to hash.** `ps -eo pid,pcpu,pmem,rss,state,comm` gives the executable name alone, so "name plus `argv[0]`" could not be built from what existed. The collector grows a `Command` field, filled from a second `ps` call joined on PID rather than an extra column — `comm` can itself contain spaces, which is why the existing parser joins the tail of the line, so appending `args` to the same output leaves no unambiguous split.

**`network` is a LAN ping sweep, not interfaces.** Tracking devices appearing on the network is a better feature than the one I described, and it does not belong in `report`: `Scan` sweeps every address in the subnet before reading ARP, which is seconds of work on a command people run in cron. That half is not in this PR and needs its own issue, where the question that decides its shape can be answered on its own.

**The threshold is measured, not chosen.** Two runs thirty seconds apart on an idle machine reported `new head` and `new sed` — the shell pipeline that was reading the report. Anything under a minute old is now ignored, and after that change forty idle seconds reported one line, for a process that had genuinely been running. A process held back is not lost; it appears on the next run, by which time it has earned the word "new".

**The invocation is hashed and never written down.** Command lines carry secrets in flags and `~/.homebutler/reports/snapshots/` has never had to be handled as a credential store. `Command` and `Elapsed` are both `json:"-"`, so neither the `processes` command nor its MCP tool gains a field. `TestSnapshotNeverHoldsACommandLine` plants a token in an invocation and asserts it is absent from the marshalled snapshot.

**Snapshot size, since the issue asked for a measurement rather than a guess.** About 50 KB on a desktop macOS with 639 distinct identities, against 2 KB before; a Linux server running a handful of services is well under that. At the default `--keep 30` that is roughly 1.5 MB of history in the worst case measured. No cap: the number did not call for one, and it is in `docs/report.md` next to `--keep` so the next person deciding does not have to re-derive it.

Verified by running it, not only by tests: baseline, forty seconds, report — and separately with a process started and stopped between runs. Snapshots written during that were removed afterwards. `gofmt`, `go vet`, `go build` and `go test ./...` are clean; `golangci-lint` is not installed here, so CI runs it first.

The report card in the README does not gain a process row here. It should, along with the README lede and the repository description, when the release that makes the whole claim true is cut — not one asset at a time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+**`report` observed processes on every run and remembered none of them.** The snapshot held containers, ports and disks; `processes` was collected, rendered, and thrown away, so "what started running since yesterday" had no path to an answer.
+
+### ✨ Features
+
+- track processes across runs (#59). A process is identified by its executable name and a hash of its full invocation, so a name that appears, disappears, or comes back running something else is reported — `python3 /opt/old.py` becoming `python3 /opt/new.py` is a `replaced` rather than silence. Identical invocations collapse to one identity, because eight workers sharing a command line are one thing running and a pool resizing is a resource signal rather than an arrival
+- ignore processes that have not been running for a minute. This threshold was measured rather than chosen: two runs thirty seconds apart on an idle machine reported `new head` and `new sed`, which was the shell pipeline reading the report. A process held back is reported on the next run, by which time it has earned the word
+
+### 🔐 Security
+
+- keep the executable name and a twelve-character hash of the invocation, never the command line itself. Command lines carry secrets in flags and `~/.homebutler/reports/snapshots/` has never had to be handled as a credential store. `ProcessInfo.Command` is excluded from JSON, so neither the `processes` command nor its MCP tool gains the field
+
+### ⚠️ Behavior changes
+
+- **Snapshots are larger.** On a desktop macOS with 639 distinct process identities a snapshot is about 50 KB, against 2 KB before; a Linux server running a few services is well under that. At the default `--keep 30` that is roughly 1.5 MB of history in the worst case measured. No cap is imposed — the measurement did not call for one — but it is worth knowing before turning `--keep` up
+
+## [Unreleased]
+
 **`report` compared counts, so a container replaced by a different container read as no change.** The snapshot on disk held the full container and port lists all along; the diff read four integers off it and threw the lists away. Six running before and six running after was reported as "No significant changes since last report", whether or not the six were the same six.
 
 ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,24 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
-**`report` observed processes on every run and remembered none of them.** The snapshot held containers, ports and disks; `processes` was collected, rendered, and thrown away, so "what started running since yesterday" had no path to an answer.
-
-### ✨ Features
-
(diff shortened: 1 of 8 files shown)


Higangssh:
Pushed fixes for six problems found re-reading this. Two were false reports.

**The first run after upgrading claimed every process on the machine had started.** A snapshot written before this change has no `processes` key and no `processes` entry in `failed_collectors`, so the guard passed and `diffProcesses(nil, …)` read the whole machine as arrivals — 639 lines on the desktop this was measured on. Verified against the real pre-existing snapshots on this machine, which now produce:

```
   skipped  processes  not compared — the previous snapshot has none
```

**`fillCommands` ran inside `allProcesses`**, so every caller of `TopProcesses` and `ListProcesses` paid for a second full `ps` sweep for two fields excluded from JSON — the TUI every two seconds, the dashboard on every request — and the report path called it twice and discarded the first result. It runs once now, in `AllProcesses`.

Also: `replaced` is reserved for one invocation before and one after, since `python3` and `ssh` routinely run several and one of them exiting is not a replacement; a process the invocation lookup missed has an unknown age rather than a young one, so a daemon that missed the join no longer disappears and comes back; the process collector is opt-in rather than part of `DefaultCollectFuncs`, because `inventory scan`, the MCP tools and `doctor` were paying for a sweep they never render and `doctor --strict` could have failed CI over a collector it does not check; and CHANGELOG had grown a second `[Unreleased]` heading.

Each has a test named after the symptom.

PR #146, opened 2026-09-05
feat(dashboard): show Proxmox freshness states

### LINKED ISSUE #106: dashboard: show Proxmox freshness and stale-data state

gsaraiva2109:
## What does this PR do?

- Adds a server-stamped freshness envelope to the existing Proxmox status response.
- Keeps the latest readable snapshot per endpoint and marks it stale after a total refresh failure.
- Exposes safe failure classes per collector without parsing or displaying credentials.
- Renders current, partially readable, stale, unavailable, and not configured states with accessible status text.
- Adds frontend state coverage and runs it in CI.

## Verification

- `go test ./... -race` with an isolated HOME
- `go vet ./...`
- `golangci-lint run`
- `go build ./...`
- `npm --prefix web test`
- `npm --prefix web run build`
- `npm --prefix web audit`

## Checklist

- [x] `gofmt -w .` - code is formatted
- [x] `go build ./...` - builds without errors
- [x] `go test ./...` - all tests pass
- [x] README updated for the user-facing change

Closes #106


Higangssh (changes requested):
This closes #106 against its own "Done when" list, item by item, and the part I expected to be fudged is the part that is right.

The issue asks for empty-but-readable data to stay distinct from ACL-filtered data, and #95 had already decided an empty resources response *is* a permissions problem — which reads like a contradiction. It is not, because the two states are reachable separately: a cluster with nodes and no VMs answers `current` and renders `No guests found`, while an ACL that hides everything leaves all three lists empty, `DefaultView` marks the collector failed, and the card renders `Node inventory unavailable`. Two tests, named after exactly that difference. That is a subtle thing to get right and it would have been easy to collapse.

The server also stops handing raw error text to the browser:

```go
- writeError(w, ..., fmt.Sprintf("get Proxmox status from %q: %v", endpoint.Name, err))
+ s.writeProxmoxFailure(w, endpoint.Name, proxmox.Classify(err), nil, nil)
```

#95 established that shape for the credential path; the status path had been leaking `%v` since. The new dependencies are all `devDependencies`, so nothing reaches the binary, and `make build-web` runs before the test step so CI has what it needs.

**Two things before merging, and the first ships to every user.**

**The dashboard now runs Svelte's development build.** `resolve.conditions: ['browser']` *replaces* Vite's defaults rather than adding to them, so `production` is lost and the exports map resolves the dev entry. Built both ways on this branch:

| | bundle | `svelte.dev/e/` strings | `dev_stack` |
| --- | --- | --- | --- |
| this PR | 86.33 kB | 12 | 1 |
| `conditions: ['module', 'browser', 'production']` | 75.38 kB | 1 | 0 |

Eleven kilobytes on top of the ~9 the card genuinely adds, and dev-mode assertions running in everyone's browser. Listing the full set rather than the one you need keeps vitest working — I ran the suite with it and all seven pass — so this does not need to be conditional on the test environment.

**The colour that says the data is old never applies.** `class:unavailable` is on the freshness line, and in the compiled CSS:

```css
.unavailable.svelte-1jhwvik,.warnings.svelte-1jhwvik{color:var(--yellow)}
...
.freshness.svelte-1jhwvik{color:var(--text-secondary)}
```

Same specificity, `.freshness` declared later, so it wins on every card. A stale panel is the same muted grey as a current one, which removes the at-a-glance signal the feature exists to add — the text still says it, but "stale data can look healthy" is the sentence the issue opens with.

**Where the operator lost information**

- When all three collectors fail, `httpFailureClass` maps only 401 and 403, so a Proxmox answering 500 — or a reverse proxy returning HTML — produces `{"status":"unavailable","failed_collectors":[...]}` with no class and no text, and the card shows `Unavailable · Failed collectors: version, cluster, resources`. Before this PR that case returned the actual error. Same for `proxmox.New`'s unclassified failures (`token ID and token are required`, `invalid port`): the card now says the single word `Unavailable` where it used to say what was wrong. The classification is the right direction — this is about what to do with what it cannot classify, and dropping to silence is worse than the generic sentence it replaced.
- `writeProxmoxFailure` takes `failureClasses` but uses it only in the `stale` branch; the `unavailable` branch drops the map, so the card falls back to the single top-level class for every collector. An unclassified decode failure on `version` next to two 403s gets reported as `version: Authorization/ACL failure`, which blames an ACL for a parse error.

**Smaller**

- The `readable` fixture has no `warnings`, but the server appends one for every failed collector, so a realistic `partially_readable` payload renders two `role="status"` elements and `findByRole('status')` throws on "multiple elements". The partial and stale tests never see the shape the server actually sends.
- `formatTimestamp` replaces `.000Z`, which matches roughly one millisecond in a thousand; the tooltip shows `…56.789Z` the rest of the time.
- `/api/proxmox/status` no longer returns 500 or 502 — every failure is 200 with a body. That is an API contract change for anything curling or health-checking it, and belongs under `⚠️ Behavior changes` rather than inside the feature bullet.

`gofmt`, `go vet`, the `server` and `proxmox` tests, `npm test` and `npm run build` all pass on `3e8716e`; CI is green.

Thanks for finishing the set. Five issues, five PRs, and this one had the most room to quietly not do what the issue asked — the empty-versus-filtered distinction especially.

gsaraiva2109:
Addressed the review in `c396561`:

- restored Vite conditions to `module`, `browser`, and `production`; the production bundle is back to 75.49 kB with no `dev_stack`
- made the unavailable freshness color win in compiled CSS
- preserved per-collector failure classes for unavailable responses and stopped falling back to the top-level class when a collector is unclassified
- added a safe generic message for unclassified failures without returning raw Proxmox errors
- made the partial/stale fixtures include server warnings and handle both status regions
- normalized fractional-second timestamps and documented the HTTP 200 freshness contract under Behavior changes
- rebased onto current `main` and preserved the new #147 changelog entry

Validated with `go mod tidy -diff`, `go vet ./...`, `go build ./...`, `go test ./... -race`, `golangci-lint run`, `npm --prefix web test`, and `npm --prefix web run build`. All six hosted checks are green.

Higangssh (approved):
All of it, and I re-measured the two that were measurable rather than taking the numbers.

The bundle is back to a production build:

| | bundle | `svelte.dev/e/` | `dev_stack` |
| --- | --- | --- | --- |
| previous head | 86.33 kB | 12 | 1 |
| `c396561` | 75.49 kB | 1 | 0 |

and in the compiled CSS `.freshness` now comes before `.unavailable,.warnings`, so the yellow wins and a stale card looks different from a current one. That was the finding I cared most about — the issue opens with "stale data can look healthy", and the colour was the only part of the answer that a glance catches.

The `unavailable` branch carries `FailureClasses` now, so a collector with an unclassified error is no longer painted with a neighbour's class. The fixtures carry `warnings`, which is the shape the server actually sends. `formatTimestamp` matches any millisecond value rather than the one-in-a-thousand that ended in `.000`. The HTTP contract change is under `⚠️ Behavior changes` where someone health-checking the endpoint will find it.

`gofmt`, `go vet`, the `server`, `proxmox` and `doctor` tests, and the seven frontend tests all pass locally on `c396561`; CI is green.

**One thing I am not holding this for**, and it is not yours.

For an unclassified failure the card now reads:

```
Unavailable · Proxmox status is unavailable
```

which is no longer silence, but it is the status field restated. The reason it cannot say more is that the error is unclassified in the first place: `proxmox.New` returns bare errors for `proxmox host is required`, `token ID and token are required`, `invalid port`, and `invalid timeout`, so `Classify` has nothing to give you. Sending the raw text instead would undo what #95 settled.

That is a question about `internal/proxmox`'s vocabulary rather than about this PR — I have now bumped into it three times, in #120, in #123, and here — and 1.0 freezes the answer. I am opening it separately.

Thanks for finishing the set. Five issues, five PRs, and this one had the most room to quietly not do what the issue asked: the empty-versus-ACL-filtered distinction is the part I expected to be collapsed and it is the part that is most carefully right.

PR #151, opened 2026-09-05
fix: classify Proxmox configuration failures

### LINKED ISSUE #149: proxmox: a config mistake has no failure class, so every consumer invents one
## Description

`proxmox.New` returns bare errors for the things an operator gets wrong in a config file:

```go
proxmox host is required
proxmox token ID and token are required
invalid Proxmox port %d
invalid Proxmox timeout %s
```

Only the TLS path is wrapped with `WithFailureClass`. So `Classify` returns `""` for every one of them, and each consumer of the classification has to decide what to do with nothing.

They have each decided differently, and none of the answers is good:

- **The dashboard** (#146) cannot send the raw error — #95 settled that — so it sends a generic message and the card reads `Unavailable · Proxmox status is unavailable`, which is the status field restated.
- **`doctor`** (#123) falls through `proxmoxFailureAction` to `"Run homebutler proxmox status for the full error"` — the diagnosis command telling the operator to go run a different command.
- **`watch`** (#124) files it under `FailureTransport`, which is the default arm of its switch, so a missing token is reported as a network problem.

## What it is not

This is not a request to send the error text to the browser. The classification exists so that callers can branch without reading messages, and #95's reasoning about what reaches an unauthenticated dashboard still holds.

## The question

What class does a configuration mistake have? The two candidates:

- **A new class**, `FailureConfiguration` or similar. Honest, and every consumer gets something it can act on: doctor can say "the endpoint is misconfigured" and name the field, the dashboard can say "not configured" rather than "unavailable", watch can stop calling it transport.
- **No class, handled by position.** An error out of `New` is a construction failure by definition — the caller knows where it came from without asking. Cheaper, and it argues the consumers should stop calling `Classify` on construction errors at all.

I have bumped into this three times now — #120 first, where I said it was worth revisiting when doctor needed to tell them apart; #123, where doctor did and the answer was to move CA and fingerprint errors back to `FailureTLS`; and #146, where the dashboard could only say nothing.

## Why it should be settled before 1.0

1.0 freezes the MCP tool surface and the JSON schema, and `failure_class` now reaches the browser through `/api/proxmox/status`. Adding a class afterwards is additive; changing what an existing one means is not.

## Notes

`internal/config`'s `ResolveCredential` already wraps its failures as `FailureAuthentication` (#125), so a missing token file is classified and a missing token in the config file is not — the same operator mistake landing on two sides of the line.

gsaraiva2109:
## What does this PR do?

Adds `FailureConfiguration` for Proxmox settings rejected before dialing and `FailureResponse` for endpoints that return invalid Proxmox API payloads.

It also keeps peer verification errors under `FailureTLS`, gives doctor actionable guidance, preserves exact classes in watch incidents, and shows missing dashboard configuration without exposing raw errors.

Closes #149

## Testing

- `go test ./... -race` with an isolated HOME
- `go vet ./...`
- `golangci-lint run`
- `go build ./...`
- `npm --prefix web test`
- `npm --prefix web run build`
- `git diff --check`

## Checklist

- [x] `gofmt -w .` code is formatted
- [x] `go build ./...` builds without errors
- [x] `go test ./...` all tests pass
- [x] User-visible behavior documented in `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### ✨ Features
+
+- add `configuration` and `response` Proxmox failure classes (#149). The dashboard distinguishes an unconfigured endpoint, doctor names the configuration fields to check, and watch records the precise class
+
 ## [0.28.0](https://github.com/Higangssh/homebutler/compare/v0.27.0...v0.28.0) - 2026-09-05
 **`doctor` knew six things it never said, and the dashboard could show you yesterday's cluster without mentioning that it was yesterday's.** Nothing was installed to poll the watch list; notifications were configured and switched off; a config holding plaintext secrets was readable by everyone; a container held the Docker socket, which is host root wearing a container's clothes; an endpoint had been left accepting any certificate after a debugging session; incident history was one restart from discarding the oldest. Every one of those was already in data homebutler had.
 
@@ -13,7 +19,6 @@ All notable changes to this project will be documented in this file.
 ```
 
 ### ✨ Features
-
 - report a watch list that will tell nobody (#147). Three gates stand between an incident and a message and two are closed by default — `watch.notify.enabled` is false, and `notify_on: flapping` means a single restart is not sent — while `doctor` only checked the third, whether a channel exists at all. The person that failed is the one who did the work: configured Telegram, ran `notify test`, saw it arrive, installed the service, and was covered except for a flag nobody mentioned. When notifications are on, `doctor` now also states what the current `notify_on` will and will not send, because the value alone does not tell an operator which incidents reach them
 - show whether each Proxmox dashboard snapshot is current, partial, stale, unavailable, or not configured (#106). Refresh timestamps now come from the server with the data, partial collectors keep readable results visible, and a failed refresh retains the last readable snapshot with only its safe failure class
 - report a running container that mounts the Docker socket (#99). A container with `/var/run/docker.sock` mounted can create containers on the host as root — it holds host root however unprivileged it looks — and homebutler installs one itself, since `install portainer` mounts the socket and nothing has mentioned it since. Costs one `docker inspect` per running container and stops at the first collector failure rather than retrying per container
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -217,7 +217,7 @@ func (p ProxmoxConfig) TokenFilePath() string {
 func (p ProxmoxConfig) TokenValue() (string, error) {
 	if p.TokenFile == "" {
 		if p.Token == "" {
-			return "", fmt.Errorf("proxmox token is not configured")
+			return "", proxmox.WithFailureClass(proxmox.FailureConfiguration, fmt.Errorf("proxmox token is not configured"))
 		}
 		return p.Token, nil
 	}
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -249,8 +249,8 @@ func TestProxmoxConfigTokenValue(t *testing.T) {
 	if got, err := inline.TokenValue(); err != nil || got != "inline-token" {
 		t.Errorf("TokenValue() = (%q, %v), want (inline-token, nil)", got, err)
 	}
-	if _, err := (ProxmoxConfig{}).TokenValue(); err == nil {
-		t.Error("TokenValue() should fail without a token source")
+	if _, err := (ProxmoxConfig{}).TokenValue(); err == nil || proxmox.Classify(err) != proxmox.FailureConfiguration {
+		t.Errorf("TokenValue() error = %v, want configuration error without a token source", err)
 	}
 
 	if _, err := (ProxmoxConfig{TokenFile: filepath.Join(dir, "missing"), Token: "inline-secret"}).TokenValue(); err == nil {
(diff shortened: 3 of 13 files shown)


Higangssh (changes requested):
This implements the decision on #149 faithfully and goes past it in two places that were right to take. `FailureConfiguration` covers what `New` refuses before it dials, `FailureTLS` keeps what fails while verifying a peer, and the line between them is the one I described. Classifying the `x509.ParseCertificate` failure inside `VerifyPeerCertificate` closes the last runtime TLS path that was leaking out as transport, and `FailureResponse` is a better answer for the decode errors than folding them into an existing class.

The part that matters is that all three consumers actually changed. `doctor` names the fields instead of telling the operator to run a different command — I ran it against `port: 99999` on both branches:

```
before  → Run homebutler proxmox status for the full error.
after   → Check the endpoint configuration fields: host, port, timeout, token ID and token, fingerprint, and CA file.
```

`watch` gives both classes their own arm rather than letting them fall into the transport default, and the card knows both labels. Go tests on the four touched packages pass, and the eleven frontend tests pass.

**One change before merging, and the wording that caused it is mine.**

I wrote on #149 that the dashboard could say "not configured" rather than restating "unavailable". I was thinking of a missing token. The class covers five things, and four of them *are* configured:

```svelte
{:else if data.status === 'unavailable' && data.failure_class === 'configuration'}
  Proxmox is not configured
```

An operator whose `ca_file` is rotated out from under them used to see `Unavailable · TLS verification failed`. They now see `Proxmox is not configured` — and line 180 explicitly suppresses the label that would have explained it. That string is byte-identical to line 163, which is what the card shows when the config has no Proxmox endpoints at all, so the two states cannot be told apart in the DOM.

`Unavailable · Configuration failed` is the honest rendering for the whole class, and the label already exists — it is only reachable through the `stale` branch today, so the same root cause is worded two ways depending on whether a snapshot happens to be cached. Sorry for pointing you at the narrower phrasing.

**Second: the `FailureResponse` action names the case it does not cover.**

"Check that the configured host and port serve the Proxmox API, not a reverse-proxy error page" is exactly right for a reverse proxy answering — and a reverse proxy answering does not get this class. `FailureResponse` is set only where a 2xx body fails to decode; a 502 with an HTML error page goes through `httpFailureClass`, which returns `""` for everything but 401 and 403. So nginx up with `pveproxy` down still produces the generic "run a different command" advice, which is the case this PR exists to remove.

Mapping unexpected statuses to `FailureResponse` would make the guidance match the scenario it names. That is the smaller half of the fix and it is where the class earns its keep.

**Smaller**

- `ResolveCredential(false)`'s "token is not configured" moved to `FailureConfiguration`, but its analogue on the action path is still `FailureAuthentication` and `ActionTokenValue`'s "action token is not configured" is still unclassified. An operator who set `action_token_id` and no token gets the authentication advice — "the token may be missing, unreadable, or revoked" — for a configuration mistake.
- `DefaultView.fail` documents that a later classified error outranks an earlier unclassified one. Classifying decode failures changes which error wins: a garbled `/version` body next to two 403s now keeps the decode error as `FirstErr`, so doctor points at the host serving the API rather than at PVEAuditor. Narrow, but the comment describing the rule was not revisited.
- No `⚠️ Behavior changes` section. An unreadable `ca_file` or malformed `fingerprint` moves from `tls` to `configuration` in `Incident.ProxmoxClass`, in the notification text, and in `/api/proxmox/status`. Anything grepping incident history or alerts for `tls` loses those matches on upgrade, and that section is the reason someone reads a changelog before upgrading.
- `docs/proxmox.md` still lists the watch-integration causes as "TLS, authentication/token expiry, or transport/timeout". `internal/watch/store.go`'s comment was updated in this PR; the page users read was not.
- The diff removes a blank line inside the released 0.28.0 section, so it no longer matches what was tagged. Worth restoring — I deleted a released section outright yesterday resolving a conflict here, so shipped content drifting by one line is a thing I would rather catch early.

`gofmt`, `go vet`, the `proxmox`, `doctor`, `server` and `watch` tests, and `npm test` all pass on `3e79fdd`; CI is green.

Thanks for taking this within the hour, and for adding `FailureResponse` rather than filing decode failures under something that already existed.

Higangssh (approved):
**A correction on my last review first.** I wrote that the checks passed "on `3e79fdd`". No such commit exists in this PR — the head I reviewed was `b547cdf`. The verification was real; the identifier I attached to it was not, and a claim about what was checked is worth nothing if it names the wrong thing.

Everything from that review is addressed, and I ran the two blockers rather than reading them.

The card no longer special-cases the class. A configuration failure renders `Unavailable · Configuration failed`, which is distinguishable from `Proxmox is not configured` — that string now appears only where it is true, when the config has no endpoints — and the label is no longer suppressed, so the same root cause is worded the same way whether or not a snapshot is cached.

`httpFailureClass` defaults to `FailureResponse`, so nginx up with `pveproxy` down finally gets the guidance that was written for it. That was the half where the class earns its keep, and the action text now names a case it actually covers.

The action-credential path is `FailureConfiguration` on both arms, so an operator who set `action_token_id` and no token stops being told their token may have been revoked. `fail`'s comment was reworded to describe what it now does rather than the rule it used to follow. `docs/proxmox.md` lists the two new causes. The `⚠️ Behavior changes` entry names the `tls` → `configuration` move, which is the one thing here that changes a value somebody may already be matching on.

And the blank line inside the released 0.28.0 section is back — I checked that section against its tag and it is byte-identical again.

Verified on `70d6cf7`: `gofmt`, `go vet`, and the `proxmox`, `doctor`, `server`, `config` and `watch` tests, plus the eleven frontend tests. `doctor` against `port: 99999` prints the field list rather than "run a different command". CI is green.

Six PRs now, and the pattern has held every time: the second version answers the review and the third answers the part of the review that was wrong. Thanks — this one closes a question I had bumped into three times without settling.

Issue #154, opened 2026-09-07
serve: the dashboard shows every setting and can change none of them


Higangssh:
## Description

The dashboard reads every setting it needs and changes none of them. Across `routes()` there is exactly one endpoint that is not a GET:

```go
s.mux.HandleFunc("POST /api/wake/{name}", api(s.handleWakeSend))
```

and that sends a magic packet, not a config change. So the operator watching memory sit at 84% against an 85% threshold has to leave the page, get to a shell on that machine, and edit YAML — to change a number the page is already showing them.

## What it costs

The thresholds are the sharpest case because the dashboard is where you learn the number was wrong, and the one place you cannot fix it. Wake targets are the same shape: a MAC is typed once, wrong, and the failure shows up in the UI that cannot correct it.

## Scope

Depends on a config writer (#153) — this issue is the surface on top of it.

- `PUT /api/config/alerts` for the three thresholds, and add/edit/remove for wake targets.
- `watch.notify` settings, which are the other thing operators change after seeing behaviour rather than before.
- Servers and Proxmox endpoints stay read-only here. They carry credentials and a trust decision, and half-editing them is worse than not offering it — the Config tab should say so where it currently says `Edit via CLI: homebutler init`.
- Surface the server's validation error in the form rather than duplicating the rules client-side. There is one validator and it should stay that way.
- While `routes()` is open, assert `/` through the mux. Since #155 the frontend tests call `frontendHandler` directly, so deleting `s.mux.Handle("/", frontendHandler(webFS))` leaves the suite green — the write endpoints are registered in the same function, and a route that nothing asserts is the one that goes missing in a rewrite.

## Notes

`--token` is optional today, and the only nudge toward it fires when the bind address is public:

```go
if ports.IsPublicBind(s.host) {
    ... "Consider using --token to require authentication."
}
```

For a read-only dashboard on `127.0.0.1` that is a defensible default. For one that can rewrite the config it is not: the same page would let anything that can reach the port change a wake target or an alert threshold. Writes should require a token whether the bind is public or not, so that turning on editing cannot quietly widen what an unauthenticated request can do.

Higangssh:
Everything this issue asked for is on `main`.

| Part | Where |
| --- | --- |
| A config writer that edits lines rather than rewriting the file | #184 |
| Alert thresholds editable, with a revision so an edit made in an editor is not discarded | #185 |
| Notification channels: a form per channel, credentials write-only, and a test send that reports each channel | #186 |
[...]

PR #155, opened 2026-09-07
fix: serve showed a directory listing when the dashboard was never built

### LINKED ISSUE #152: serve: the dashboard is a file listing when the web assets were never built
## Description

`serve` embeds `internal/server/web_dist`, which is gitignored apart from `.gitkeep`. `make build-web` fills it, and CI and the release workflow both run it. `go build` and `go install` do not.

`frontendHandler` decides whether the dashboard was built by counting entries:

```go
entries, err := webFS.ReadDir("web_dist")
if err != nil || len(entries) == 0 {
    // fallbackHTML: "Web dashboard not built yet" / make build-web
}
```

`.gitkeep` is tracked, so the count is never zero and the fallback is unreachable. What answers `/` instead is `http.FileServer` over a directory with no `index.html` — its autoindex:

```
$ go build -o hb . && ./hb serve --port 8787
homebutler dashboard: http://127.0.0.1:8787

$ curl -s http://127.0.0.1:8787/
<!doctype html>
<meta name="viewport" content="width=device-width">
<pre>
<a href=".gitkeep">.gitkeep</a>
</pre>
```

Checked on darwin/arm64 at 2772a4a. With `web_dist` populated everything else about `serve` holds up: the SPA loads, unknown paths fall back to `index.html`, assets are served, and `/api/status?server=rpi5` routes over SSH and comes back with the remote host.

## What it costs

The README documents `go install github.com/Higangssh/homebutler@latest`. Anyone who takes that route gets a working CLI and a dashboard that is a file listing of one dotfile — no error, no explanation, and the page written for exactly this situation never renders.

The fallback also names `make build-web`, which is not available to the person most likely to see it: no checkout, no Makefile, no npm. Reached or not, it points at the wrong repair.

`TestSPAFallback_UnknownPath` passes today because it accepts anything with a 200 — "index.html (or fallback HTML)" — so the suite cannot tell the dashboard from an autoindex.

## Scope

- Decide "built" by opening `web_dist/index.html`, not by counting entries.
- Rewrite `fallbackHTML` for both audiences: from a checkout, `make build-web`; installed as a binary, the release download ships the dashboard already.
- Assert the distinction in a test rather than accepting any 200 — the frontend `fs.FS` has to be injectable for that to be testable at all.

7487:
## What does this PR do?

Fixes #152.

`serve` decided the dashboard was built by counting `web_dist` entries. The tracked `.gitkeep` made that count never zero, so `go install` and a plain `go build` produced a binary whose `/` was `http.FileServer`'s autoindex of one dotfile, and the fallback page written for exactly that case never rendered. "Built" now means `web_dist/index.html` opens.

The fallback page speaks to both audiences. Someone who came in through `go install` has no checkout, so it points at the release binaries and `install.sh`, which ship the dashboard. From a checkout it says `make build-all` rather than the previous `make build-web`: the dashboard is embedded at compile time, so building the web assets without rebuilding the binary leaves the running one exactly as it was.

`frontendHandler` takes the `fs.FS` as a parameter instead of reading the package-level embed, which is what makes the two cases testable with `fstest.MapFS`. `TestFrontendFallback` now asserts the fallback page byte for byte from a `web_dist` holding only `.gitkeep`, and `TestSPAFallback_UnknownPath` asserts the body is `index.html` rather than accepting any 200. The unreachable `fs.Sub` warning is folded into the same fallback branch.

Left alone: the README's "Build from Source" says `make build`, which produces the same dashboard-less binary. Happy to change that to `make build-all` here if you want it in the same PR.

Verified on darwin/arm64: `go build -o hb . && ./hb serve --port 8787`, `curl -s http://127.0.0.1:8787/` returns the fallback page with this change and the `.gitkeep` autoindex without it.

## Checklist

- [ ] README updated (if user-facing change)


Higangssh (commented):
The fix is the right predicate. `web_dist` is a build artefact whose only tracked file is `.gitkeep`, so any test of "is there anything in here" was going to answer yes forever. `index.html` is the file that decides whether there is a dashboard to serve, and it is the one you check. Taking the `fs.FS` as a parameter instead of reading the package-level embed is what makes either branch testable at all, and it is the change I would have asked for if it had not already been here.

Nothing blocking. Verified rather than read:

- `gofmt -l`, `go vet ./...`, `go build ./...` and `go test ./...` are clean on the branch. `golangci-lint` is not installed on this machine, so CI is the check for that one.
- Built the branch with `web_dist` holding only `.gitkeep`: `/` returns the fallback page. The same build answers `<a href=".gitkeep">.gitkeep</a>` without the change.
- Copied a real `dist/` in and rebuilt: `/` is the dashboard, `/nope` falls back to `index.html`, and `/assets/` returns `index.html` rather than a listing, because `sub.Open("assets/")` fails on the trailing slash and the SPA branch catches it. Worth saying explicitly — the bug being fixed was a directory listing, so it is fair to ask whether any path still produces one. None does.
- `TestFrontendFallback` fails against the old handler, which is what makes it a regression test rather than a description of current behaviour.

`make build-all` over `make build-web` is correct, and it is the detail I expected to have to point out. The dashboard is embedded at compile time, so rebuilding the assets without rebuilding the binary leaves the running one exactly as it was.

Two minor things, take them or leave them.

**The README offer: yes, take it in this PR.** `### Build from Source` ends in `make build`, which produces the same dashboard-less binary through the one route where the reader definitely does have a Makefile. Fixing the fallback text while leaving the instruction that leads people to it is half the change.

**Nothing asserts the mux still routes `/`.** Both tests now call `frontendHandler` directly, so deleting `s.mux.Handle("/", frontendHandler(webFS))` from `routes()` would leave the suite green. Going through `srv.Handler()` is what the old `TestSPAFallback_UnknownPath` bought in exchange for being unable to assert a body, since the embedded FS differs between a checkout and CI. One line asserting that `/` through the mux is not an autoindex would keep both.

The dropped `log.Printf` on the `fs.Sub` failure is fine by me. It is unreachable with a constant path, and the only cost is that in the impossible case the page would explain the wrong thing.

Thanks for this one. Taking the `fs.FS` as a parameter and asserting the fallback byte for byte are the two decisions that make it hold up, and both were easy to skip.

Higangssh:
Merged. On the README line you offered to take: I merged before you had a chance to push it, so I have taken it myself in #156 rather than leave you holding a one-line follow-up. `make build-all` there, with the Node requirement stated, since `make build` not needing it is a real reason someone would pick it.

The mux-level assertion I mentioned is still open and still not urgent. If you want it, it belongs in whatever you look at next rather than a PR of its own.

7487:
Thanks — saw #156, and stating the Node requirement next to `make build-all` is the right call since skipping it is the one reason to pick `make build`. Noted on the mux-level assertion: I'll fold it into the next change I touch in `serve` rather than open a PR for one line.

Higangssh:
For the record rather than to reopen anything: I put the mux-level assertion in the scope of #154 instead of an issue of its own. That work registers the write endpoints in `routes()`, so it is the change where the route wiring is already in front of whoever is doing it — and a one-line test does not need its own issue, PR and CI run to be remembered.

Issue #136, opened 2026-09-03
report: Needs Attention is blind to everything the diff just found


Higangssh:
## Description

`report` has two sections that answer different questions and never speak to each other.

`Notable Changes` now knows what moved, by identity. `Needs Attention` is still filled entirely from thresholds on the *current* reading — memory over 85%, a disk over 85%, a count of stopped containers (`internal/report/report.go`, the "Needs attention" block). Not one entry comes from the diff.

So this is what a report looks like when a service quietly became reachable from the whole network:

```
── Notable Changes ───────────────────────────────────────────
   gone  127.0.0.1:8080/tcp  was vaultwarden
   new   :8080/tcp           now vaultwarden
   disk  /                   +2.4 GB since last report

── Needs Attention ───────────────────────────────────────────
   (nothing)
```

Two lines of equal weight, and the one that matters is indistinguishable from the one that means it is Tuesday.

## What "worth mentioning" is missing

The README says homebutler decides what is worth saying. It currently decides what earns *a line* — that is the suppression rule in `docs/report.md`, and it works. It does not decide what is worth *acting on*, which is the other half and the one a person opens the report for.

## Scope

Derive `Needs Attention` entries from the changes, alongside the existing threshold checks:

- **A listener that became publicly reachable.** Either a new public listener, or the same port and protocol moving from a specific address to a wildcard. The second is invisible today: it renders as a `gone` and a `new` and nothing correlates them.
- **A container that is not running after being replaced or re-imaged.** A redeploy that did not come back is the failure a person most needs told, and today it is a `replaced` line whose detail happens to end in `running → exited`.
- **Containers that stopped, by name.** The existing entry says "3 container(s) stopped" while the diff knows which three.

## Notes

`needs_attention` is a `[]string` field and 1.0 freezes the JSON schema, so the shape of what goes in it should be settled before then. Adding entries is additive; changing what the section means later is not.

Related: the suggested actions have the same gap (#137), and the ordering does too (#138).

Issue #133, opened 2026-09-03
mcp: the unsupported-version error lists the version it just refused


Higangssh:
## Description

`UnsupportedProtocolVersionError` lists every version this server implements, including the one it just refused:

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{"_meta":{
    "io.modelcontextprotocol/protocolVersion":"2024-11-05",
    "io.modelcontextprotocol/clientCapabilities":{}}}}' | homebutler mcp

{"error":{"code":-32022,"message":"Unsupported protocol version",
  "data":{"requested":"2024-11-05",
          "supported":["2026-07-28","2025-11-25","2025-06-18","2025-03-26","2024-11-05"]}}}
```

Both halves are individually true. homebutler does implement `2024-11-05` — through `initialize` — and the [versioning page](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning) says the error lists "the versions it does support". Put together they contradict each other: the client is told a version is unsupported and handed a list containing it.

## What breaks

The spec's instruction to the client is:

> The client **SHOULD** select a mutually supported version from the `supported` list and retry the request

A client that picks `2024-11-05` from that list retries with the same `_meta` and gets the same error. Nothing in the protocol stops it choosing the oldest, and nothing here tells it not to.

## Why it is not urgent

It takes a modern client sending a legacy version in `_meta` to reach at all, which is already an odd thing for a modern client to do. Nobody is hitting this.

## Why it should still be settled

1.0 freezes the MCP surface, and what a dual-era server advertises in this error is part of it. Deciding after 1.0 means deciding it as a breaking change.

## The distinction that resolves it

The two lists answer different questions and should stop being the same variable:

- `server/discover` advertises **what this server can be reached by**, both eras. All five is right, and the spec's own example mixes eras.
- `-32022` answers **what may appear in `_meta`**, which by the revision's own terminology is the modern versions only.

`isModernProtocolVersion` already exists from #128 and draws exactly this line; the error payload is the one place still using the combined list.

## Notes

Found while reviewing #128, which is otherwise complete. Not that PR's to fix — the question predates it and deserves its own answer.

PR #174, opened 2026-09-12
watch: parse older incident filenames for prune/doctor

### LINKED ISSUE #169: watch: retention counts one set of incidents and prunes another
## Description

Two functions count the same directory and disagree, and the disagreement decides what gets deleted.

`ListIncidentRefs` parses the filename and skips anything that does not fit the current format:

```go
// The suffix is 6 hex characters and the timestamp is fixed width, so the
// container name is whatever precedes them, dashes and all.
const tsLayout = "20060102-150405.000"
```

`ListIncidents` opens every `.json` file and keeps whatever unmarshals. On this machine that is 5 incidents by one count and 4 by the other:

```
crash-test-20260410-224033.581-7a2124.json   parsed
crash-test-20260410-224039.904-83fbc6.json   parsed
crash-test-20260410-224049.429-21ea13.json   parsed
crash-test-20260410-224105.370-b090a0.json   parsed
ghostmeet-backend-1-20260410-174933.json     skipped — no milliseconds, no suffix
```

The skip is deliberate for the reason the comment gives: a name homebutler cannot read is not one it should delete. But two things are built on the count rather than on the skip.

## What it costs

`PruneIncidents` works from the refs, so a file in the older format is never a candidate for deletion. `max_incidents` bounds what it can see, and a directory holding older incidents exceeds the limit by exactly the number it cannot parse, silently and permanently. #101 added that bound so incident history would stop growing forever; this is the case where it does not.

`doctor` reports "N of M incidents kept" from the same refs, so the number an operator is shown is the smaller one. The warning fires late, and on a directory made entirely of older files it never fires at all.

This was found while building the Watch view (#165), which needed a count that agreed with the list underneath it and got `watch.CountIncidents` instead.

## Scope

Decide what an unparseable name means, once, and have all three agree:

- Widen `parseIncidentRef` to read the older format as well. It is a real format homebutler wrote, so refusing to read it is refusing its own history. Then prune and doctor cover everything and `CountIncidents` collapses back into `ListIncidentRefs`.
- Or keep the refusal and make it visible: prune counts every `.json` file toward the limit while still only deleting what it can parse, and `doctor` says how many it could not read rather than leaving them out of the total.

The second is worse but honest; the first is what the directory's own history argues for. What is not acceptable is the current state, where the limit reports one number, enforces another, and nobody is told which.

team-humaki:
Fixes #169.

Older incident files looked like `gitea-20260410-174933.json`. `parseIncidentRef` only knew the ms+suffix form, so prune/doctor skipped them and the retention count drifted from the list.

Now both name shapes parse, and `CountIncidents` uses the same refs.


Higangssh (commented):
This takes the option #169 argued for rather than the one that was easier to land, and the details that would have made it subtly wrong are the ones you got right.

Ordering the current format first and falling through matters more than it looks: `ghostmeet-backend-1-20260410-174933.json` has five dash-separated parts, so the current-format branch reaches for `1-20260410` as its timestamp, fails, and hands over. That file is in the table, which is the case a naive split would have quietly mangled into a container called `ghostmeet-backend`. The guarantee that a name homebutler cannot read is never deleted survives intact — the final `return IncidentRef{}, false` is still the only way out for an unrecognised name. And `CountIncidents` collapsing back into the refs is exactly what should happen once the root cause is fixed; it existed only because the two disagreed.

**Before merge: this needs a `⚠️ Behavior changes` entry in `CHANGELOG.md`.** It changes what gets deleted, and it does so on data the operator cannot get back.

- `internal/watch/store.go:225` — `PruneIncidents` bounded only what it could name. The half of #169 that reads as a bug is that `max_incidents` under-enforced; the half that reads as a feature is that those files were therefore immortal. After this they are pruning candidates, so someone upgrading with April-era incidents loses the oldest of them on the next `SaveIncident`, without having asked for anything. That is the case the section exists for.
- `internal/doctor` picks the change up for free, which is the point, but it means the number in `N of M incidents kept` goes up on an existing directory. A `--strict` cron that was passing can start warning on a machine nobody touched. Worth the same entry rather than a separate one.

Two minor things, take them or leave them.

`CountIncidents` is now `len(refs)` and has one caller. Keeping it is fine and arguably clearer at the call site; folding it away is equally fine. That one is mine to decide, not yours — I added it in #170 as the temporary half of this, and you have removed the reason it existed.

The new `PruneIncidents` comment says "any incident format homebutler has written". True today, and it is the sentence that goes stale first if a third shape ever appears. `parseIncidentRef` names both forms explicitly a few lines down; matching that would age better.

Verified rather than read: `gofmt`, `go vet ./...`, `golangci-lint run` (0 issues), `go build ./...`, and the `watch`, `server` and `doctor` packages all pass on the branch. I also pointed a build at the real watch directory on this machine, which happens to be the mixed case — four current-format incidents and one `ghostmeet-backend-1-20260410-174933.json` — and the refs now name all five. The dashboard's count does not move, because #170 had it counting `.json` files directly; that was the workaround and this removes the need for it.

Thanks for picking this one up. #169 was written as "decide what an unparseable name means, once", and reading its own older format rather than legislating around it is the answer that leaves the least behind.

team-humaki:
Added the `⚠️ Behavior changes` entry (older-format files become pruning candidates; doctor counts can rise on an untouched directory) plus a matching Fixes line, and tightened the `PruneIncidents` comment to name both formats the way `parseIncidentRef` does.

Higangssh:
That is the entry, and the second bullet is one I had not asked for — a green `--strict` cron starting to warn on a machine nobody touched is the half of this someone finds out about at 3 AM.

One thing left, and it is the only one: the Fixes line cites `(#174)`, which is this pull request. Entries in this file cite the issue instead — `#102`, `#105`, `#111`, `#133`, `#136` are all issues — because the number is what a reader follows to find out why the change was argued for, and the issue is where that argument is. Yours is `#169`.

Change that one reference and I will merge. Everything else is done: CI is green including the end-to-end job, and I verified the code and the mixed-format directory locally before the review above.

team-humaki:
Switched the changelog cite to #169.

Higangssh:
Merged. Three round trips and each one landed exactly what it was asked for, which is not the usual experience of a first contribution to someone else's codebase.

Two things worth saying back. The second behaviour change bullet — a green `--strict` cron starting to warn on an untouched machine — was yours, not mine, and it is the one an operator meets without warning at an inconvenient hour. And you took the harder half of #169: reading the format homebutler used to write, rather than legislating around it, which is what leaves nothing behind for the next person to trip over.

`CountIncidents` was mine and existed only because the two counts disagreed; it is now one line over the refs, and folding it away entirely is a tidy-up I will do when something else takes me through that file.

Your next PR will not need the workflow approved by hand — that gate only applies until a first one is merged.

